PR #86, opened 2021-08-16
[F] Build loading progress bar

jen-castiron:
NOTE: This currently feels a little hacky. The FullPageLoader is styled to be at the top of the page no matter where the component is added. This is because the loading state is inside QueryWrapper, which isn't necessarily at the top of the page. A little javascript sets the FullPageLoader's position based on the top of the `<main>` tag. A solution that might be a little less hacky could be a context that has an isLoading var that can be set from anywhere. This is how loading is handled in OFB food finder.


zdavis:
I think what you describe is the right solution! The Query Wrapper could trigger the loading state in context without rendering the visual loading indicator.

zdavis:
Looks good to me! +1

PR #98, opened 2021-08-28
[B] Fix pagination

jen-castiron:
The problem:
Pagination and ordering are not working as expected. Both should be controlled using URL query parameters, so that links to a specific page and order can be shared. Currently clicking on next page or an order does nothing.

This PR fixes the issue by getting the query params from the router and passing them to QueryWrapper. There needs to be a little further refactoring to prevent the page from reloading everything (header, search bar, etc) when getting new items.

BUT! I noticed there is a `setVariables` method getting passed from QueryWrapper to its children. It's currently not used in any lists. 

I want to make sure I'm not stepping on anyone's toes here before I make further changes. Are we all on the same page for using the URL query to search, sort, and paginate the list?


zdavis:
Rebased on main to incorporate minor next.js config fix.

PR #101, opened 2021-08-30
[F] Create links input repeater, add to contributor forms

jen-castiron:
This is most of the way there - it's an input repeater specifically for Contributors. I figure once we get this to where we want it, it can be abstracted for other repeaters. 

WIP, but needs review for type issues and general structure. 

- [x] Finish styling
- [x] Default values are not working as expected

Storybook component:
https://wdp-storybook-b2124k6lm-cast-iron.vercel.app/?path=/story/components-composed-contributor-contributorlinksinput--in-a-form


components/composed/contributor/ContributorUpdatePersonForm/ContributorUpdatePersonForm.tsx
```diff
@@ -49,6 +50,7 @@ export default function ContributorUpdatePersonForm({ data }: Props) {
       mutation={mutation}
       getErrors={getErrors}
       toVariables={toVariables}
+      defaultValues={{ links }}
```
jen-castiron, line 53: This is a bit frustrating... in order to get the correct default value to useFieldArray, the default must be set at the form level. (You cannot set a default value at the field array level.)

jen-castiron: The way the default value is set on other forms might be unnecessary, and it's possible that we should be passing the default value to the parent form, _not_ the child inputs.

components/composed/contributor/ContributorUpdatePersonForm/ContributorUpdatePersonForm.tsx
```diff
@@ -45,6 +71,7 @@ export default function ContributorUpdatePersonForm({ data }: Props) {
       mutation={mutation}
       getErrors={getErrors}
       toVariables={toVariables}
+      defaultValues={defaultValues}
```
jen-castiron, line 74: Hey @scryptmouse ! I was able to solve my type issue in ContributorLinksInput, but I'm getting an odd type error with the default values here. Error: https://vercel.com/cast-iron/wdp-admin-client/x9V7EpXW41rGEyPPTV7JKtbcT7TZ
```
Type error: Type '{ givenName?: string \| null \| undefined; familyName?: string \| null \| undefined; title?: string \| null \| undefined; email?: string \| null \| undefined; affiliation?: string \| null \| undefined; bio?: string \| ... 1 more ... \| undefined; links?: readonly ({ ...; } \| null)[] \| undefined; " $refType": "ContributorUpdateP...' is not assignable to type '{ title?: string \| null \| undefined; url?: string \| null \| undefined; email?: string \| null \| undefined; links?: { title?: string \| undefined; url?: string \| undefined; }[] \| null \| undefined; ... 5 more ...; clearImage?: boolean \| ... 1 more ... \| undefined; }'.
--
```
The error popped up after I moved the query vars out of the shared field fragment

components/composed/contributor/ContributorUpdatePersonForm/ContributorUpdatePersonForm.tsx
```diff
@@ -63,7 +90,24 @@ const mutation = graphql`
   ) {
     updatePersonContributor(input: $input) {
       contributor {
-        ...ContributorPersonFormFragment
+        givenName
```
zdavis, line 93: @scryptmouse, @jen-castiron  — Is there no way to avoid repeating the exact same set of fields in our update forms?

jen-castiron: I'm not sure, so I gave this a try:
```
export default function ContributorUpdatePersonForm({ data }: Props) {
  const { contributorId } = useFragment(fragment, data);

  const { image, ...defaultValues } = useFragment(fieldsFragment, data);
 
  ...
}

...

const fieldsFragment = graphql`
  fragment ContributorUpdatePersonFormFieldsFragment on AnyContributor {
    ... on PersonContributor {
      givenName
      familyName
      title
      email
      affiliation
      bio
      image {
        thumb {
          png {
            alt
            url
          }
        }
      }
      links {
        title
        url
      }
    }
  }
`;

const mutation = graphql`
  mutation ContributorUpdatePersonFormMutation(
    $input: UpdatePersonContributorInput!
  ) {
    updatePersonContributor(input: $input) {
      contributor {
        ...ContributorUpdatePersonFormFieldsFragment
      }
      ...MutationForm_mutationErrors
    }
  }
`;

const fragment = graphql`
  fragment ContributorUpdatePersonFormFragment on AnyContributor {
    ... on PersonContributor {
      contributorId: id
      ...ContributorUpdatePersonFormFieldsFragment
    }
  }
`;
```

But relay insists that ContributorUpdatePersonFormFieldsFragment isn't spread, and seems to be confusing the second fields fragment for the first query fragment.
```
Invariant Violation: Relay: Expected to receive an object where `...ContributorUpdatePersonFormFieldsFragment` was spread, but the fragment reference was not found`. This is most likely the result of:
- Forgetting to spread `ContributorUpdatePersonFormFieldsFragment` in `useFragment()`'s parent's fragment.
- Conditionally fetching `ContributorUpdatePersonFormFieldsFragment` but unconditionally passing a fragment reference prop to `useFragment()`. If the parent fragment only fetches the fragment conditionally - with e.g. `@include`, `@skip`, or inside a `... on SomeType { }` spread  - then the fragment reference will not exist. In this case, pass `null` if the conditions for evaluating the fragment are not met (e.g. if the `@include(if)` value is false.)
- 
```

jen-castiron: Of course, I could be thinking about this the wrong way and trying to get the fields from the wrong set of data.

jen-castiron: Of course, right after I typed that I think I got it! Changes pushed!

zdavis: Did you try this?

```
const contributor = useFragment(fragment, data);
const { image, ...defaultValues } = useFragment(fieldsFragment, contributor)
```

I think that using the fragment in the mutation would require changes to MutationForm. I'm not sure what MutationForm does with the response from the mutation, if anything. Now that I'm thinking about this more, do we even need to specify the fields in the mutation at all?

scryptmouse: `fieldsFragment` can go directly on a `PersonContributor`, no need to put it on the union. I think that might be mucking things up with the update form part:

```graphql
fragment ContributorUpdatePersonFormFieldsFragment on PersonContributor {
  fieldsGoHere
}

fragment ContributorUpdatePersonFormFragment on AnyContributor {
  ... on PersonContributor {
    ... ContributorUpdatePersonFormFieldsFragment
  }
}
```

I'm not sure exactly where the last error is coming from, but in my experience that error is from passing the wrong "level" of data to the useFragment hook, like, for the mutation, you'll need to do `useFragment(fieldsFragment, response.updatePersonContributor.contributor)`—the object passed to useFragment needs to be the exact object the fragment was spread on

jen-castiron: I did try that, but I had to add a type to `useFragment` to get it to pass linting. On if we need the fields in the mutation... I'm not sure! A good question for @scryptmouse

scryptmouse: The fields in the mutation form are for passing to onSuccess to do something with them. THey don't have to be the same fields though.

scryptmouse: Though it can also be good to retrieve the fields, even if we don't grab them with `useFragment`, in order to update the Relay cache

scryptmouse: I wonder if the union is a little weird with Relay fragments directly like that. Every time I've used relay frags with unions, it's always been with dispatching on `__typename`—like I wonder if typescript is expecting us to check to make sure `contributor.__typename === "PersonContributor"` first.

I could pretty easily add a top-level `personContributor` and `organizationContributor` resolver if we want

scryptmouse: Or, maybe we just need a dispatch above the update forms:

```graphql
fragment UpdateAnyContributorFormFragment on AnyContributor {
  __typename
  ... on PersonContributor {
    ... UpdatePersonContributorFormFragment

  ... on OrganizationContributor {
    ... UpdateOrganizationContributorFormFragment
  }
}
```

```typescript
export default function UpdateAnyContributorForm(({ data }) {
  const contributor = useFragment(fragment, data);

  if (!contributor) {
    return null;
  } else if (contributor.__typename === "PersonContributor") {
    return <UpdatePersonContributorForm data={contributor} />;
  } else if (contributor.__typename === "OrganizationContributor") {
    return <UpdateOrganizationContributorForm data={contributor} />;
  }
}
```

zdavis: Yep, that's what we're currently doing in `ContributorUpdateForm`:

```
  if (contributor.__typename === "PersonContributor")
    return <ContributorUpdatePersonForm data={contributor} />;
  if (contributor.__typename === "OrganizationContributor")
    return <ContributorUpdateOrganizationForm data={contributor} />;
```

zdavis: It's good to know that these could be added at the top level though — there are a few places where dealing with the %other type or ensuring you have the right type is a little annoying. I don't know that we need it here, but we may want that at some point.

components/composed/contributor/ContributorCreatePersonForm/ContributorCreatePersonForm.tsx
```diff
@@ -21,8 +22,26 @@ export default function ContributorCreatePersonForm() {
...
+        <Forms.FileUpload label="Image" name="image" />
+        <Forms.Input label="Title" {...register("title")} />
+        <Forms.Email
+          label="Email"
+          {...register("email")}
+          description="Format: example@email.com"
+        />
+        <Forms.Input label="Affiliation" {...register("affiliation")} />
+        <Forms.Textarea label="Bio" {...register("bio")} />
+        <LinksRepeater<Fields>
```
jen-castiron, line 38: Last bit... now that it's abstracted, this should probably be `Forms.LinksRepeater`. I'm just not sure where that's defined.

PR #105, opened 2021-09-02
[WIP][B] Fix excess page rendering

jen-castiron:
This is a start to fixing excess page re-rendering.

The problem: QueryWrapper determines what should show if the query is loading, erroring, or if the data is null or empty. This results in blank pages when data loads.

The solution: The children of QueryWrapper should determine what to do with load states, erroring, or empty/null data. 

Check out Collections to see current progress!

TODO:
- [ ] Pass loading and error states down to QueryWrapper's children
- [ ] Show table loading state instead of "no data"
- [ ] Investigate if there's a better solution to null data in ModelPageCountActions
- [ ] Update other list pages to match Collections


zdavis:
I forked this branch and used it as the starting point for #108, which takes Jen's approach here and applies it across the board, with some updates/fixes based on discussions that Alexa and I had while Jen was out last week. I'm closing this and we can move it forward in #108.

PR #119, opened 2021-09-13
[B] Fix tab active state in sub routes

jen-castiron:
- Resolves WDP-189


components/layout/PageHeader/PageHeader.tsx
```diff
@@ -26,7 +26,9 @@ const PageHeader = ({ title, breadcrumbsProps, tabRoutes, buttons }: Props) => {
             {tabRoutes.map(({ label, ...namedLinkProps }, i) => (
               // router and activeRoute may be missing in Storybook
               <NamedLink key={i} {...namedLinkProps} passHref>
-                <TabNav.Tab active={activeRoute?.name === namedLinkProps.route}>
+                <TabNav.Tab
+                  active={activeRoute?.name.includes(namedLinkProps.route)}
```
jen-castiron, line 30: When we define child routes in `baseRoutes.ts`, the child route's name should include it's parent. For example:
parent: `collection.manage`
child: `collection.manage.details`
In PageHeader's case, we know the tabs should be active if the child route is active.

PR #125, opened 2021-09-13
[F] Add subHead prop to PageHeader and parents

1aurend:
I've added a boolean prop `subHead` to the `PageHeader` component and a pass-thru for every list and layout component that eventually calls it.

Questions:
1. Is the bool a good approach or is it better to pass a string like "h2" so there's more flexibility?
2. I had to add the `.a-hidden` to the whole `Styled.Header` to get the spacing right when the title isn't visible. Is there ever a case where the breadcrumbs would need to be visible but the title not?

### DIFF
--- a/components/composed/collection/CollectionList/CollectionList.tsx
+++ b/components/composed/collection/CollectionList/CollectionList.tsx
@@ -12,6 +12,7 @@ import type { ModelTableActionProps } from "react-table";
 
 function CollectionList<T extends OperationType>({
   data,
+  subHead,
 }: CollectionListProps) {
   const destroy = useDestroyer();
   const drawerHelper = useDrawerHelper();
@@ -48,12 +49,14 @@ function CollectionList<T extends OperationType>({
       actions={actions}
       selectable
       data={collections}
+      subHead={subHead}
     />
   );
 }
 
 interface CollectionListProps {
   data?: CollectionListFragment$key;
+  subHead?: boolean;
 }
 
 type CollectionNode = CollectionListFragment["nodes"][number];
--- a/components/composed/collection/CollectionList/CollectionList.tsx
+++ b/components/composed/collection/CollectionList/CollectionList.tsx
@@ -12,7 +12,8 @@ import type { ModelTableActionProps } from "react-table";
 
 function CollectionList<T extends OperationType>({
   data,
-  subHead,
+  headerStyle,
+  hideHeader,
 }: CollectionListProps) {
   const destroy = useDestroyer();
   const drawerHelper = useDrawerHelper();
@@ -49,14 +50,16 @@ function CollectionList<T extends OperationType>({
       actions={actions}
       selectable
       data={collections}
-      subHead={subHead}
+      headerStyle={headerStyle}
+      hideHeader={hideHeader}
     />
   );
 }
 
 interface CollectionListProps {
   data?: CollectionListFragment$key;
-  subHead?: boolean;
+  headerStyle?: "primary" | "secondary";
+  hideHeader?: boolean;
 }
 
 type CollectionNode = CollectionListFragment["nodes"][number];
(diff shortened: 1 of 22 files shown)


jen-castiron (changes requested):

components/layout/PageHeader/PageHeader.tsx
```diff
@@ -10,14 +10,23 @@ type NamedLinkProps = React.ComponentProps<typeof NamedLink>;
...
+}: Props) => {
   const activeRoute = RouteHelper.activeRoute();
 
   return (
-    <Styled.Header>
+    <Styled.Header className={hideHeader ? "a-hidden" : ""}>
       {!isNil(breadcrumbsProps) ? <Breadcrumbs {...breadcrumbsProps} /> : null}
       <Styled.TitleWrapper>
-        <Styled.H1>{title || "\u00a0"}</Styled.H1>
+        <Styled.Title as={headerStyle === "primary" ? "h1" : "h2"}>
```
jen-castiron, line 27: Let's check for "secondary" and default to "h1" if headerStyle is empty

components/composed/collection/CollectionList/CollectionList.tsx
```diff
@@ -48,12 +50,16 @@ function CollectionList<T extends OperationType>({
...
+      headerStyle={headerStyle}
+      hideHeader={hideHeader}
     />
   );
 }
 
 interface CollectionListProps {
   data?: CollectionListFragment$key;
+  headerStyle?: "primary" | "secondary";
+  hideHeader?: boolean;
```
jen-castiron, line 62: Let's DRY the list type definitions by extending from PageHeader's types:
```
// Get PageHeader's types:
type HeaderProps = React.ComponentProps<typeof PageHeader>;

...

// Extend CollectionListProps with PageHeader:
// Pick lets us choose which props we want to use
interface CollectionListProps extends Pick<PageHeader, 'headerStyle' | 'hideHeader'> {
    data?: CollectionListFragment$key;
}
```

Now if the definition of `headerStyle` or `hideHeader` changes in PageHeader, we won't have to update the other components

jen-castiron (changes requested):
Almost there! I've added a few comments where things can be fixed / DRYed up. 

The padding is a bit off on the subheaders - it looks like we can remove the top padding if the style = "secondary", so the header lines up with the first sidebar nav item:
Screen Shot 2021-09-13 at 2 24 28 PM

components/composed/collection/CollectionList/CollectionList.tsx
1aurend: @jen-castiron I should do this for each component that uses the props?

jen-castiron: @1aurend Yep - it's pretty verbose, but it makes sure that the definition isn't repeated outside of `PageHeader`

components/layout/PageHeader/PageHeader.tsx
jen-castiron: Nevermind - I missed that this has a default value 😅

components/layout/PageHeader/PageHeader.styles.ts
```diff
@@ -3,10 +3,18 @@ import { basePadding } from "theme/mixins/appearance";
...
 
-export const Header = styled.header`
-  padding-block-start: ${basePadding(6)};
+export const Header = styled.header<Pick<Props, "headerStyle">>`
+  ${({ headerStyle }) =>
+    headerStyle !== "secondary" &&
+    css`
+      padding-block-start: ${basePadding(6)};
+    `}
+
```
jen-castiron, line 17: Here's how I fixed the header padding! First the header prop types needed to be imported (line 6 & 7), then the prop for headerStyle needed to be added to the styled header (line 11). After that, you can use the headerStyle value to add top padding if the style isn't "secondary" ✨

jen-castiron (approved):

PR #126, opened 2021-09-13
[B] Add labels to error messages and required fields

jen-castiron:
- Resolves WDP-183

### DIFF
--- a/components/forms/Errors/MessageList.tsx
+++ b/components/forms/Errors/MessageList.tsx
@@ -3,7 +3,7 @@ import type { MultipleFieldErrors, ValidateResult } from "react-hook-form";
 
 import Message from "./Message";
 
-export default function MessageList({ inlineMessage, messages, label }: Props) {
+export default function MessageList({ inlineMessage, messages }: Props) {
   if (!messages) {
     if (inlineMessage) {
       return <Message result={inlineMessage} />;
@@ -15,21 +15,19 @@ export default function MessageList({ inlineMessage, messages, label }: Props) {
   return (
     <>
       {Object.entries(messages).map(([type, result]) =>
-        renderMessage(type, result, label)
+        renderMessage(type, result)
       )}
     </>
   );
 }
 
-function renderMessage(type: string, result: ValidateResult, label: string) {
+function renderMessage(type: string, result: ValidateResult) {
   return result !== null ? (
-    <Message key={type} type={type} result={`${label} ${result}`} />
+    <Message key={type} type={type} result={result} />
   ) : null;
 }
 
 interface Props {
   messages?: MultipleFieldErrors;
   inlineMessage?: string;
-  /* The label for the erroring field, ie 'First Name' */
-  label: string;
 }
--- a/components/forms/Errors/MessageList.tsx
+++ b/components/forms/Errors/MessageList.tsx
@@ -3,7 +3,7 @@ import type { MultipleFieldErrors, ValidateResult } from "react-hook-form";
 
 import Message from "./Message";
 
-export default function MessageList({ inlineMessage, messages }: Props) {
+export default function MessageList({ inlineMessage, messages, label }: Props) {
   if (!messages) {
     if (inlineMessage) {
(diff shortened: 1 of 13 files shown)


components/forms/Errors/MessageList.tsx
```diff
@@ -15,21 +15,21 @@ export default function MessageList({ inlineMessage, messages }: Props) {
...
-  return <Message key={type} type={type} result={result} />;
+function renderMessage(type: string, result: ValidateResult, label: string) {
+  return result !== null ? (
+    <Message key={type} type={type} result={`${label} ${result}`} />
+  ) : null;
 }
 
 interface Props {
   messages?: MultipleFieldErrors;
   inlineMessage?: string;
```
jen-castiron, line 32: @scryptmouse What does an example inlineMessage look like? Would it need the field label as well?

scryptmouse: I'm not 100% sure, it gets used if the errors just have one message associated with them, which I can think can happen with some manual RHF validations—it's see line 17 in `components/forms/Errors/Errors.tsx`. It may be safe to omit with how we're generating errors and pretty much exclusively setting them on the server.

jen-castiron:
@zdavis & @scryptmouse , this PR adds the label to field error messages. An alternative to adding the label  would be to write error message that are label agnostic. For example, take a look at how HTML5 returns error messages:
Screen Shot 2021-09-13 at 2 00 45 PM
Screen Shot 2021-09-14 at 12 27 12 PM
Screen Shot 2021-09-14 at 12 28 07 PM
Do we want to go forward with adding labels to each message, or use language that is label agnostic?

zdavis:
Really good question/suggestion. I never thought about solving this error messaging problem in this way. I'm certainly open to trying it, as I think it's actually pretty challenging to reliably combine field name and validation error message. I'm totally open to taking this approach.

zdavis:
One other thought — we might just want the API to return something more like `error.blank` or `error.invalidEmail` that we can pass to our localization system. Are we really going to want to localize these message on the server?

scryptmouse:
I think the server should be responsible for localizing its errors—a single source of truth vs needing to have localization files in every consumer of the API, and also, I think the server needs the flexibility to add new types of errors without the client knowing about them.

(also, maybe a biased opinion, but Ruby's i18n format & support still seems a lot more refined than what exists in JS land, especially when handling plurals and such)

zdavis:
Fine! :)

Can the server provide error messages with wording like what Jen is proposing?

scryptmouse:
To make sure I'm following it entirely, we're wanting full messages (with the localized attribute name) vs the predicate format I was returning previously?

Or are wanting the HTML5-style?

The former is trivial, it's just a matter of calling `full_message` instead of `message`, though are we okay with duplication in the (semi-rare, but not impossible) instance that an attribute has more than one error?

zdavis:
In that case, I guess I don't really care, but lean slightly toward HTML style error wording because it's simpler. It always kind of bugs me to see the field name repeated when there are multiple errors.

jen-castiron:
I agree with Zach - it feels repetitive to include the field name in each individual error. The component is written in such a way that the error should be accessible, even if it doesn't include the field name. But while it seems simpler to provide the same message for certain errors, I don't know how complicated it is on the backend.

jen-castiron:
@zdavis @scryptmouse - I've removed the FE label from error messages so that the errors can be:
1. handled from the backend
2. label agnostic (HTML5 style labels) if possible
Do you want me to make a new issue for the backend work?

scryptmouse:
I also agree about not duplicating the field names. Restructuring the errors to read more like HTML5 is very doable—the strings don't really care how they're formatted, so long as they're consistent, and there aren't that many of them:

https://github.com/NGLPteam/wdp-api/blob/main/config/locales/en.yml are the custom ones, and there are a couple other files in the same directory that I plucked out of the gems we use for validating API inputs with the expectations that we'd adjust wordings.

And a new issue would be great, thank you!

PR #128, opened 2021-09-13
[WIP] Add size prop to ButtonControl

jen-castiron:
Resolves WDP-182


components/atomic/buttons/ButtonControl/ButtonControl.styles.ts
```diff
@@ -21,20 +21,18 @@ export const ButtonControl = styled.button<BaseProps>`
   opacity: var(--button-control-opacity, 1);
   visibility: var(--button-control-visibility, visible);
 
-  > span {
-    display: inline-block;
-  }
+  ${({ size }) =>
+    size === "large" &&
```
jen-castiron, line 25: @1aurend The biggest change from your branch was setting size to "large" instead of "42px". When we're creating style based props for components, we should give the style a name instead of passing a specific value to the CSS. This way we can:
1. Ensure button sizing is consistent
2. Reduce the cognitive load on devs ("large" is easier to remember than exact pixel size)
3. Change the size across the app, if needed

The pattern is similar to how we would define the large version of a button with SASS BEM rules:
```
// a button module
.m-button { 
   ... 
   // styles for the 'large' version
   &--large { ... } 
} 
```

components/atomic/buttons/ButtonControl/ButtonControl.styles.ts
```diff
@@ -65,7 +63,18 @@ export const ButtonControl = styled.button<BaseProps>`
...
-  padding: ${pxToRem("7px")};
-  min-width: auto;
+export const ButtonText = styled.span<Pick<BaseProps, "size" | "icon">>`
+  display: inline-block;
+
+  ${({ size }) =>
+    size === "large" &&
+    css`
+      padding-inline-start: ${pxToRem("4px")};
+    `}
```
jen-castiron, line 73: In this block, I had to adjust the text's padding for large buttons. Lael's design called for slightly more padding at the start of a button if there's text.

PR #131, opened 2021-09-14
[F] Add secondary headers to layouts

zdavis:
@jen-castiron and @1aurend — I went ahead and implemented these secondary headers in the layouts, building on your changes to PageHeader. Please feel free to adjust the approach in this PR in a subsequent commit if you're not happy with it. I just wanted these looking good for the demo tomorrow, and it only took a few minutes to get them in place.


zdavis:
One issue is with the "hideHeader" flag. I'm hiding the header on the contribution lists because I want the layout to handle the header. In this case, I think it's probably wrong to include the header for screen readers, so we may need a little more control over when it's rendered as well as when it's visible.

PR #132, opened 2021-09-14
[C] DRY PageHeader props

1aurend:
@jen-castiron I think these props are correct now, except `ModelListPage`. Which type there do I extend to add the header props? I have this, which I don't think is right:

```
type ModelListPageProps<
  T extends OperationType,
  U extends PaginatedConnectionish,
  V extends Record<string, unknown> = Record<string, unknown>
> = Omit<ModelListProps<T, U, V>, "view"> &
  Pick<ModelListActionsProps, "viewOptions"> &
  Pick<HeaderProps, "headerStyle" | "hideHeader"> & {
    defaultView?: ModelListActionsProps["active"];
    buttons?: ReactNode;
  };
```

### DIFF
--- a/components/composed/model/ModelListPage/ModelListPage.tsx
+++ b/components/composed/model/ModelListPage/ModelListPage.tsx
@@ -12,6 +12,9 @@ import type { ModelListActionsProps } from "components/composed/model/ModelListA
 import { QueryVariablesContext } from "contexts";
 import { OperationType } from "relay-runtime";
 import { useIsMobile } from "hooks";
+import PageHeader from "components/layout/PageHeader";
+
+type HeaderProps = React.ComponentProps<typeof PageHeader>;
 
 export type PaginatedConnectionish = Connectionish &
   ModelPaginationFragment$key &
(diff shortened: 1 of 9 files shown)


components/composed/model/ModelListPage/ModelListPage.tsx
```diff
@@ -22,11 +25,10 @@ type ModelListPageProps<
   U extends PaginatedConnectionish,
   V extends Record<string, unknown> = Record<string, unknown>
 > = Omit<ModelListProps<T, U, V>, "view"> &
-  Pick<ModelListActionsProps, "viewOptions"> & {
+  Pick<ModelListActionsProps, "viewOptions"> &
+  Pick<HeaderProps, "headerStyle" | "hideHeader"> & {
```
jen-castiron, line 29: This looks right to me - we're extending a type here, while the other components use an interface. An interface is mutable, and we just want to add two new properties to what's already defined.

Here's more info on the difference between using `interface extends` and `type &`:
https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

jen-castiron (approved):
👍 looks good!

1aurend:
Thanks @jen-castiron! I do need to spend some more time in the TS docs. What should happen with the branch conflicts? Is it okay to resolve the conflicts by keeping all the imports and the nameColumn prop? Or is there a better way?

jen-castiron:
@1aurend I took a quick look at the conflicts, and that sounds about right. You'll want to rebase our branch on main, then resolve the conflicts:
```
git pull --rebase origin main
```
It will tell you there's conflicts, and that's when you can see which files have issues:
```
git status
...
Unmerged paths:
  (use "git restore --staged <file>..." to unstage)
  (use "git add <file>..." to mark resolution)
	both modified:   components/composed/contribution/CollectionContributionList/CollectionContributionList.tsx
	both modified:   components/composed/contribution/ItemContributionList/ItemContributionList.tsx
```
Open those files and fix the issue, then add all changes and continue:
```
git add .
git rebase --continue
```
Once all conflicts are fixed, force push your branch:
```
git push -f
```
Then (hopefully) all checks will have passed and you'll be able to merge your changes 👍

PR #135, opened 2021-09-14
[F] Refactor PageHeader and ContentHeader layout components 

jen-castiron:
There is a lot going on here, but the main idea is that we _do not know_ if page headers will be plain, need room for buttons, or even need to be on the page at all (in the case of lists). This is what makes it so difficult to abstract a layout, and can make abstracting layouts actually more cumbersome. Comments below.

WDP-208

NOTE: This is a WIP. Some padding was lost while refactoring PageHeader / ContentHeader, and will be added back in.


pages/collections/[slug]/manage/access.tsx
```diff
@@ -18,7 +18,7 @@ function CollectionAccess() {
       initialVariables={{ ...queryVars, collectionSlug }}
     >
       {({ data }) => (
-        <CollectionLayout showSidebar data={data?.collection}>
+        <CollectionLayout showSidebar data={data?.collection} useRouteHeader>
```
Comment (jen-castiron) [inline pages/collections/[slug]/manage/access.tsx:21]: ## Child page headers
The problem:
Child page headers can be one of the following:
1. Plain text
2. Text and right hand content (buttons)
3. Included with a model list

Instead of hiding the model list's header, I've created this flag, `useRouteHeader`, to use when a plain text header is needed.

pages/collections/[slug]/manage/contributions.tsx
```diff
@@ -21,7 +21,6 @@ function CollectionContributions() {
       {({ data }) => (
         <CollectionLayout showSidebar data={data?.collection}>
           <CollectionContributionList<Query>
-            hideHeader={true}
```
Comment (jen-castiron) [inline pages/collections/[slug]/manage/contributions.tsx:24]: ## Child page headers
Now that we can choose what style of header to use in a child page, the header does not need to be hidden inside model lists

components/layout/ContentHeader/ContentHeader.styles.ts
```diff
@@ -0,0 +1,30 @@
+import styled, { css } from "styled-components";
+import { noFlexGapSupport } from "theme/mixins/base";
+import { pxToRem } from "theme/mixins/functions";
+import ContentHeader from "./ContentHeader";
+type Props = React.ComponentProps<typeof ContentHeader>;
+
+export const Wrapper = styled.div<Pick<Props, "headerStyle">>`
```
jen-castiron, line 7: ## ContentHeader vs PageHeader
The more I looked at these two components, the more I found a reason to keep both. The problem was that PageHeader was repeating the same styling as ContentHeader - so I've updated ContentHeader to be more versatile. In the designs, this style of header usually includes content on the right hand side. Both the title and the right hand content are underlined.

jen-castiron:
If we do want to continue to abstract layouts, this particular block keeps getting repeated on model child pages:

```
return (
    <section>
      <PageHeader
        title={<ContributorDisplayName contributor={contributor} />}
      />
      <ContentSidebar sidebarLinks={manageRoutes}>
        {activeRoute && activeRoute.label && (
          <PageHeader
        {useRouteHeader && activeRoute && activeRoute.label && (
          <ContentHeader
            headerStyle="secondary"
            title={t(`navLabels.${activeRoute.label}`)}
          />
        )}
        {children}
      </ContentSidebar>
    </section>
  );
```

I could see this getting turned into it's own Page or ModelPage component, but I'm concerned that will further complicate our app. Some pages do not follow this structure (such as the dashboard).

zdavis:
### RE: Redundant PageHeader / Sidebar / ContentHeader in the Model Layout components

I think if we try to remove the repetition now, we'd be doing it prematurely. There are a finite number of these layouts (only 5 currently), and I think it's entirely possible that with time we might seem some differences emerge in them as we get into real-world uses and need different headers, buttons, etc. I think your instinct that abstraction here would add complexity is right, and we shouldn't try to abstract these until it feels like a repeated PITA to deal with them. I'd feel differently if there was a bunch of markup or styles in these layouts, but they're just composing other components, so it doesn't worry me.

### RE: Refactoring ContentHeader and PageHeader

What you say all makes sense to me — I haven't thought through dealing with the buttons and various states as deeply as you have, so I don't have any suggestions to offer on that. At first I thought maybe we should just remove the header from the ModelLists, but there are buttons in the header that are connected to the list and its selection state, for example, so those probably belong in the ModelLists. It's complicated, and when it's complicated, we should be wary of trying to abstract too much!

### One other thought

The only thing that gave me pause in this PR really was changing the default behavior of the model layouts from "show the secondary header unless I tell you not to" to "only show it when I tell you to." That feels wrong to me, because the majority of the pages do show the secondary header—and the fact that you had to add the `useRouteHeader` prop to almost all of the pages suggests the default value for `useRouteHeader` should be `true` on the model layouts. I like the default behavior to be the most common behavior, while requiring extra configuration for deal with exceptional cases (like when there'e a list and we need to hide the header). That said, this isn't a hill I'd die on if you have a good reason for switching the default!

In short, looks good to me +1

PR #136, opened 2021-09-15
[WIP] For discussion: different layout approach

zdavis:
This PR is still a POC showing how we can improve our approach to model Layout components (CommunityLayout, CollectionLayout, etc). These layouts are used on routes where there is a `[slug]` in the URL referencing a single model. This PR shows the approach for these routes:

```
/communities/[slug]/collections
/communities/[slug]/manage/details
/communities/[slug]/manage/roles
/communities/[slug]/manage/members
```

This PR introduces a `CommunityQueryLayout` component that sits between the Page component and the QueryLayout. The page has a `getLayout` function that wraps it in `CommunityQueryLayout` and passes it its query. The `CommunityQueryLayout` component is then responsible for executing the query, passing the data to `CommunityLayout` and rendering the page with its props. Because this wrapping component is rendered in `_app`, we avoid remounting the layout. 

Not only does this approach solve our remounting problem, it DRYs up the pages. Each page can be more focused on what it's rendering and no longer has to deal with getting its slug and executing the query with QueryWrapper.

### Caveat!
You will see a flash when you navigate from `/communities/[slug]/collections` to `/communities/[slug]/manage`. This is not a deficiency with this approach. That happens because the `/communities/[slug]/manage` redirects the user to `/communities/[slug]/manage/details` so the layout is remounted. That redirect isn't really necessary — we could adjust our navigation to render the redirect path if the route is a redirect to solve this problem. This is, however, a separate issue and not something I'm solving in this PR.

So, to see the user experience improvement, navigate between details / roles / members in the sidebar nav. The flash of the title should no longer be an issue.

### DIFF
--- a/components/composed/community/CommunityLayoutQuery/CommunityLayoutQuery.tsx
+++ b/components/composed/community/CommunityLayoutQuery/CommunityLayoutQuery.tsx
@@ -1,0 +1,44 @@
+import React, { ComponentProps } from "react";
+import { QueryWrapper } from "components/api";
+import { useRouteSlug, useBaseListQueryVars } from "hooks";
+import type { QueryLayoutProps, QueryPageComponentProps } from "types/page";
+import { HasFragment } from "types/graphql-helpers";
+
+import ErrorPage from "next/error";
+import CommunityLayout from "../CommunityLayout";
+
+function CommunityLayoutQuery<
+  Query extends CommunityQuery,
+  P extends QueryPageComponentProps<Query>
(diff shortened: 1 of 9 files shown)


jen-castiron:
This looks great! And I agree with the redirect issue - we should set the manage tab's link to `manage/details`. We can leave the redirect in if someone went to `manage/`, just in case.

zdavis:
Closing and picking this up in #141

PR #137, opened 2021-09-15
[F] Add back link to pages without clear parent child relationships

jen-castiron:
After discussing the problem with the FE team, the consensus was that it's OK to have both breadcrumbs on models with parent/child relationships and a "back" link on other models. Collections and Items are easy to get back to, as their link is provided in the main nav. (While "contributors" is hidden behind a dropdown).

Having just the text "Contributors" didn't seem like enough context for the user:
Screen Shot 2021-09-15 at 9 24 24 AM

Adding an arrow and extra text seemed to help distinguish this back button from the parent/child breadcrumbs on other pages:
Screen Shot 2021-09-15 at 9 26 40 AM

There's one more problem. If I land on a contributor page from an item or collection, going "Back to Contributors" doesn't make sense. It would make more sense to say "Back to [Item Name]". Unfortunately `next/router` does not provide a way to get the last route, so we don't really know what page we're going back to. 

Questions:
1. Is the above an actual problem? If so, does a link that just says ,"Back" work, or should we keep track of the last route?
2. Should the text read "All Contributors" instead of "Back to Contributors"?
3. Other opinions?

### DIFF
--- a/components/composed/community/CommunityLayout/CommunityLayout.tsx
+++ b/components/composed/community/CommunityLayout/CommunityLayout.tsx
@@ -5,6 +5,7 @@ import { PageHeader, ContentSidebar, ContentHeader } from "components/layout";
 import { useChildRouteLinks, useMaybeFragment, useRouteSlug } from "hooks";
 import { RouteHelper } from "routes";
 import { useTranslation } from "react-i18next";
+
 export default function CommunityLayout({
   children,
   showSidebar = false,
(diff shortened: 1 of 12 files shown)


zdavis:
As I read through your description of the problem, "all contributors" is what occurred to me. I think I'd start with that and see if users get confused. I agree that the inclusion of the arrow helps considerably.

PR #142, opened 2021-09-16
[F] Create Modal layout component

1aurend:
This Modal component is based largely off Drawer. It includes a simplified header and the remainder of its content is rendered in its children. Eventually, `dialog.hide` will need to be passed to the children so buttons can close the modal.

### DIFF
--- a/components/layout/Modal/Modal.styles.ts
+++ b/components/layout/Modal/Modal.styles.ts
@@ -31,11 +31,13 @@ export const DialogBackdrop = styled(BaseDialogBackdrop)`
 export const Modal = styled(BaseDialog)`
   --modal-padding-inline: ${basePadding(10)};
   position: fixed;
-  top: 30px;
-  left: 40vw;
+  inset-block-start: 50%;
+  inset-inline-start: 50%;
+  transform: translate(-50%, -50%);
   display: flex;
   flex-direction: column;
-  max-width: ${pxToRem(497)};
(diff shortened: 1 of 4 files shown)


components/layout/Modal/Modal.styles.ts
```diff
@@ -0,0 +1,67 @@
...
+`;
+
+export const Header = styled.header`
+  padding-block-start: ${basePadding(10)};
+  padding-inline-start: var(--modal-padding-inline);
+  padding-inline-end: var(--modal-padding-inline);
+  background: var(--background-color);
+`;
+
+export const HeaderBar = styled.div`
```
jen-castiron, line 53: These two styles could probably be combined into one header style

jen-castiron (changes requested):
Getting there! A few styling issues - The modal is slightly off center - I tried removing the content to see if that was part of the issue, but it's still off center:
Screen Shot 2021-09-17 at 10 05 42 AM

The mobile version should look more like this:
Screen Shot 2021-09-17 at 10 14 53 AM

jen-castiron (approved):
@1aurend This is a nice base modal! There's a small issue with scrolling - if the modal is taller than the device, you can't scroll to see the whole modal. Check out the MobileMenu or Drawer for examples. 

Since it's late on a Friday, let's get it merged in. I can make an issue for the scroll bug. 👍

PR #144, opened 2021-09-17
[F] Add HiddenField component, only show visibility dates on 'VISIBLE'

jen-castiron:


components/composed/collection/CollectionUpdateForm/CollectionUpdateForm.tsx
```diff
@@ -56,18 +68,18 @@ export default function CollectionUpdateForm({ data, onSuccess }: Props) {
...
-        />
-        <Forms.Datepicker
-          label="forms.collection.fields.visibleAfterAt"
-          {...register("visibleAfterAt")}
-        />
-        <Forms.Datepicker
-          label="forms.collection.fields.visibleUntilAt"
-          {...register("visibleUntilAt")}
-        />
+        <Forms.HiddenField watch={watch} field="visibility" showOn="LIMITED">
```
jen-castiron, line 71: `HiddenField` takes the form's `watch` function, a field to watch, and a value to check. If the field equals the value passed in, the child fields show!

PR #145, opened 2021-09-17
[F] Improve active state in primary navigation

jen-castiron:
- Resolves WDP-228

Adds new functions to RouteHelper 
* isRouteNameActive and isRouteActive. Checks if the route is an exact match to the active route
* isRouteNameFuzzyActive and isRouteFuzzyActive. Checks  if the route or any of it's child routes are active


zdavis:
I probably won't have time to review this closely. I'm fine with it being accepted without my feedback, however.

PR #146, opened 2021-09-20
[B] Add LinksRepeater styles for small screens

1aurend:
@jen-castiron Just wanted to share this with you at this stage. What else should I do before sending to Lael for review?


components/forms/Fieldset/Fieldset.styles.ts
```diff
@@ -12,18 +12,46 @@ export const Legend = styled.legend`
   padding-block-end: var(--fieldset-field-spacing, ${basePadding(4)});
 `;
 
-export const Fields = styled.div`
+export const Fields = styled.div<Props>`
+  --table-border: 1px solid var(--border-color);
+
   display: flex;
-  flex-direction: row;
+  flex-direction: ${({ isMobile }) => (isMobile ? "column" : "row")};
```
jen-castiron, line 19: `useIsMobile` has a very rare use-case - it should only be used when a CSS media breakpoint can't be used. In this case, instead of passing a `isMobile` flag, we should use the mixin `respond` instead:
```
flex-direction: row;
...
${respond(css`
  flex-direction: column;
  border-bottom: var(--table-border);
`, 20}
```
The above example will set the direction to column and add a border when the browser's max width is the variable `20` 

The function is defined here: https://github.com/NGLPteam/wdp-admin-ui/blob/main/theme/mixins/base.ts#L8-L17
And the breakpoint values are defined here: https://github.com/NGLPteam/wdp-admin-ui/blob/main/theme/base/variables.ts#L86-L101

jen-castiron (changes requested):
You're ready for Lael to review, but I do have some suggestions on how to use our CSS  mixins for mobile layouts.

components/forms/Fieldset/Fieldset.styles.ts
1aurend: Super helpful! Thanks! ✨

jen-castiron (approved):
👏 Great work!

PR #147, opened 2021-09-20
[F] Add drawer option to ButtonControlGroup 

jen-castiron:
Created a new ButtonControl pattern called ButtonControlDrawer. As the name suggestions, this is a button control that opens a drawer! Creating a pattern made it easer to add the drawer flag to the ButtonControlGroup.

For everyone's sake, here's the difference between each control:
* ButtonControl - A styled button. Can be a link if `as='link'` prop is passed
* ButtonControlDrawer - A button control that opens a drawer. Uses DrawerLink and ButtonControl under the hood.
* ButtonControlGroup - Pass in an array of button props, and this spits out a list of button controls. On mobile, the controls are in a dropdown menu.

### DIFF
--- a/components/atomic/buttons/ButtonControl/index.ts
+++ b/components/atomic/buttons/ButtonControl/index.ts
@@ -1,3 +1,5 @@
 import ButtonControl from "./ButtonControl";
+import ButtonControlDrawer from "./patterns/ButtonControlDrawer";
 
 export default ButtonControl;
+export { ButtonControlDrawer };
--- a/components/atomic/buttons/ButtonControl/patterns/ButtonControlDrawer.stories.tsx
+++ b/components/atomic/buttons/ButtonControl/patterns/ButtonControlDrawer.stories.tsx
@@ -1,0 +1,63 @@
+import ButtonControlDrawer from "./ButtonControlDrawer";
+import { Story } from "@storybook/react";
+import { ICON_KEYS } from "components/factories/IconFactory";
+import { withDesign } from "storybook-addon-designs";
+type BaseProps = React.ComponentProps<typeof ButtonControlDrawer>;
+
+export default {
+  title: "Components/Atomic/Buttons/ButtonControlDrawer",
+  component: ButtonControlDrawer,
+  decorators: [withDesign],
+  parameters: {
+    themes: {
+      default: "neutral00",
+    },
+    design: [
+      {
+        type: "figma",
+        url:
+          "https://www.figma.com/file/EeaBT8NWvguKGhMQ7pgpry/NGLP-Admin-UI-Design?node-id=29%3A163",
+      },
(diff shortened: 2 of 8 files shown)


jen-castiron:
This seems like a pretty good pattern for issue [WDP-234, add confirmation modal to delete button in lists](https://linear.app/nglp/issue/WDP-234/add-confirmation-modal-to-delete-buttons-in-model-lists) @zdavis , let me know what you think of this plan. This PR shows how to make a reusable pattern component - a ButtonControl that opens a drawer. This idea could be used to make a confirmation or delete pattern.

The problem: We need to have a delete button that opens a confirmation modal, then runs destroy on the model.  Unfortunately, the modal cannot be opened from `useDestroyer`, since the delete button should be a disclosure with a control id that matches the modal for a11y. The modal and button can't be disconnected like they are with drawers, because drawers are a link & route and do not need a control-id.

A `ButtonControlDelete` pattern could be created, where
* Clicking the button opens a confirmation modal
* The passed in function runs on modal confirm

Example properties that would need to be passed into the new button pattern (aside from normal button props):
* Modal content
* Modal a11y title
* onDelete (or use onClick that's part of the normal ButtonControl props)

Example pattern in this PR: https://github.com/NGLPteam/wdp-admin-ui/pull/147/files#diff-277d58d4f43ad23f9056c842d6a6d03a80055ad0f934bbe36ac0d5a2251224fc

These extra properties will need to be added to buttons in ButtonControlGroup, so the right button is rendered from the list. Example from this PR: https://github.com/NGLPteam/wdp-admin-ui/pull/147/files#diff-c1de5c6e46571289a39fa05e934338e8d354b3b220e8f2c4704cc4a47ddb89be

The list buttons are defined in `useRowActions`: https://github.com/NGLPteam/wdp-admin-ui/blob/main/components/composed/model/ModelList/hooks/useRowActions.tsx

zdavis:
This looks good to me. And yes, it makes sense to me to apply the pattern to the delete problem. 

I'd recommend going with a `ButtonControlConfirm` component rather than a `ButtonControlDelete`, since the confirmation could be used for really anything that required the user to confirm, which could be actions beside deletion. Unless, that is, there are a11y requirements specifically for delete?

Every delete button in a table is currently rendered from the [useRowActions](https://github.com/NGLPteam/wdp-admin-ui/blob/main/components/composed/model/ModelList/hooks/useRowActions.tsx) hook. I like our current approach of configuring actions for the row more than an approach that allowed each table to pass in arbitrary react nodes because it ensures that the order of icons and the icon choices are always the same. Right now, these actions just take a `handleClick` property. We could expand the hook to accept the modal content and title as well, and if those properties were present, the useAction hook could render a modal button for any action, regardless of whether it was delete or something else.

PR #148, opened 2021-09-21
[F] Hide pagination controls for single pages

1aurend:
Adds a condition for hiding pagination if the total page count is 1. Also, moves the bottom Table and Grid borders into Pagination so they are hidden if the controls are hidden.


jen-castiron (approved):
👏 If it looks good to Lael, it looks good to me 🚢

zdavis:
I doubt Lael will review this unless we specifically ask him to. I don't see any issues, so I say merge it.

PR #149, opened 2021-09-21
[F] Add user avatar and dropdown menu

1aurend:
@jen-castiron I've started building the avatar dropdown, but I wanted to get your eye on it before I do more.

Things I know need to be done:
- (Likely) replace my use of `useIsAuthenticated` with `useViewerContext` to grab both the avatar image and the signin state
- Add the dropdown items to `appData` rather than the mock I'm currently using and the labels to locales
- Related, create/add the routes for "Edit Profile" and "Edit Settings"

Let me know what you think the next steps are here.

### DIFF
--- a/components/global/Header/HeaderAccount.tsx
+++ b/components/global/Header/HeaderAccount.tsx
@@ -1,12 +1,12 @@
 import React from "react";
-import SignIn from "components/scaffolding/Auth/SignIn";
-import SignOut from "components/scaffolding/Auth/SignOut";
+// import SignIn from "components/scaffolding/Auth/SignIn";
+// import SignOut from "components/scaffolding/Auth/SignOut";
 import * as Styled from "./Header.styles";
 import { Dropdown, NamedLink } from "components/atomic";
 import { Authorize } from "components/auth";
 import { useTranslation } from "react-i18next";
 import { RouteHelper } from "routes";
-import useIsAuthenticated, { useSignInOut } from "hooks/useIsAuthenticated";
+import { useSignInOut } from "hooks/useIsAuthenticated";
 type NamedLinkProps = React.ComponentProps<typeof NamedLink>;
 type AuthorizeProps = React.ComponentProps<typeof Authorize>;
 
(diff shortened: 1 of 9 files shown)


components/global/Header/HeaderAccount.tsx
```diff
@@ -1,14 +1,96 @@
...
+
+interface HeaderNavItem {
+  label: string;
+  actions?: AuthorizeProps["actions"];
+}
 
 const HeaderAccount = () => {
-  return (
-    <>
-      <SignIn />
```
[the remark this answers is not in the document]
1aurend: These components struck me as unnecessary, and I've been trying to figure out how to articulate why. I think it was in part because I learned React early in my dev journey, so it seems much more natural to conditionally render the label on a button and its onClick than to have two separate buttons that use css to hide/show. It seemed straightforward to combine the logic of both of these components into one hook and not use the components. BUT when I implemented it, it essentially converged on your idea of rolling them into a single component.

components/global/Header/HeaderAccount.tsx
```diff
@@ -1,14 +1,92 @@
...
+    return (
+      <Styled.AuthLink onClick={handleSignInOut}>
+        <Styled.Link>
+          <Styled.LinkText>
+            {isAuthenticated ? "Signout" : "SignIn"}
+          </Styled.LinkText>
+        </Styled.Link>
+      </Styled.AuthLink>
+    );
+  };
```
1aurend, line 71: Here's what could become the new combined `SignInOut` component. Also, do we use localizations for "Sign In" and "Sign Out"? I just stuck in strings as placeholders.

fixtures/app.data.ts
```diff
@@ -52,6 +52,7 @@ const appData = {
         children: [USERS_LINK, CONTRIBUTORS_LINK],
       },
     ],
+    account: { label: "account.manage", children: [] },
```
1aurend, line 55: Not sure if this is the correct label for this placeholder.

components/global/Header/HeaderAccount.tsx
jen-castiron: 👍 Nice! Yes, Sign In / Sign Out should be localized

fixtures/app.data.ts
jen-castiron: What should the label be? If it's "Manage", then use "header.manage". If it's "Account", then use "header.account". The key/values are here: https://github.com/NGLPteam/wdp-admin-ui/blob/main/locales/en.json#L144-L148
You can always add new translations to `en.json` if you need to!

1aurend:
Sorry for all the force pushes... someday I'll get the linting right... but I think this is about ready.

jen-castiron (approved):
👍 Ready to go

PR #150, opened 2021-09-22
[F] Set list view value in local storage

jen-castiron:
Leaving this open until tomorrow morning so folks have time to look at it if they'd like.

- Create shared ViewOptions enum
- Update all components to use ViewOptions enum
- Create useLocalStorage and useViewPreference hooks
- Set default list view (table) and default list view options (table only)
- Resolves WDP-132

Thank you @scryptmouse for the useLocalStorage gist!

### DIFF
--- a/components/composed/user/UserList/UserList.tsx
+++ b/components/composed/user/UserList/UserList.tsx
@@ -10,7 +10,6 @@ import { CellProps } from "react-table";
 
 import ModelListPage from "components/composed/model/ModelListPage";
 import ModelColumns from "components/composed/model/ModelColumns";
-import { DataViewOptions } from "components/atomic/DataViewToggle";
 import PageHeader from "components/layout/PageHeader";
 
 type HeaderProps = React.ComponentProps<typeof PageHeader>;
@@ -45,7 +44,6 @@ function UserList<T extends OperationType>({
     <ModelListPage<T, UserListFragment, UserNode>
       modelName="user"
       columns={columns}
-      viewOptions={[DataViewOptions.table]}
       data={users}
       headerStyle={headerStyle}
       hideHeader={hideHeader}
--- a/hooks/useViewPreference.ts
+++ b/hooks/useViewPreference.ts
@@ -1,0 +1,9 @@
+import { ViewOptions } from "utils/view-options";
+import useLocalStorage from "./useLocalStorage";
+
+/**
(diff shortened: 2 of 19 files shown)


hooks/useViewPreference.ts
```diff
@@ -0,0 +1,9 @@
+import { ViewOptions } from "utils/view-options";
+import useLocalStorage from "./useLocalStorage";
+
+/**
+ * Get and set list view preferences
+ */
+export default function useViewPreference(key = "nglp::global.listView") {
```
jen-castiron, line 7: Let me know if we have any preferences on how the key is formed

utils/view-options.ts
```diff
@@ -0,0 +1,8 @@
+export enum ViewOptions {
+  table = "table",
+  grid = "grid",
+}
+
+export const DEFAULT_VIEW_OPTIONS = [ViewOptions.table];
+
+export const ALL_VIEW_OPTIONS = [ViewOptions.table, ViewOptions.grid];
```
jen-castiron, line 8: Now we have consistent view option values between models and the atomic DataViewToggle

components/composed/user/UserList/UserList.tsx
```diff
@@ -45,7 +44,6 @@ function UserList<T extends OperationType>({
     <ModelListPage<T, UserListFragment, UserNode>
       modelName="user"
       columns={columns}
-      viewOptions={[DataViewOptions.table]}
```
jen-castiron, line 48: Because so many lists had just `table` as their view option, I've set that to be the default. It was rare to see all options used (collections, items)

zdavis:
I'd like to review this before it's merged, so please give it a little time.

jen-castiron:
One more note: If we want to detangle the model list props with the DataViewToggle, we could :
1. Pass the key to and set the view in DataViewToggle
2. [Listen for local storage changes](https://developer.mozilla.org/en-US/docs/Web/API/StorageEvent) in ModalList (where needed)

components/composed/user/UserList/UserList.tsx
zdavis: +1

zdavis:
Thanks for holding on this. +1 from me, no objections. I like how you tie the preference to the model type being displayed rather than having it be a global preference or a per-table preference. Looks great!

PR #151, opened 2021-09-22
[POC][F] Add multiselect buttons

jen-castiron:
I tried to use the same pattern as table actions for multiselect, so the action is defined in the model list. I'm using `CollectionsList` as an example here. I'm using the `ButtonControlGroup` component, because we will eventually want to have more than one action on multiple selection.

TODO:
- [ ] Check for delete permission
- Add a confirmation modal (WDP-234)
- Hook up to a multi-delete API (TBD)

Screen Shot 2021-09-22 at 11 33 32 AM

Screen Shot 2021-09-22 at 11 33 41 AM
 

Screen Shot 2021-09-22 at 11 33 56 AM


jen-castiron:
Closing - multiselect will be hidden for now

PR #157, opened 2021-09-23
[F] Create Collection Contribution form

jen-castiron:


components/atomic/buttons/ButtonControl/patterns/ButtonControlDrawer.tsx
```diff
@@ -7,9 +7,9 @@ type DrawerLinkProps = React.ComponentProps<typeof DrawerLink>;
  * A button control that opens a drawer
  */
 const ButtonControlDrawer = forwardRef(
-  ({ drawer, icon, children }: Props, ref) => {
+  ({ drawer, drawerQuery, icon, children }: Props, ref) => {
     return (
-      <DrawerLink drawer={drawer} passHref>
+      <DrawerLink drawer={drawer} drawerQuery={drawerQuery} passHref>
```
jen-castiron, line 12: While I didn't end up using drawerQuery values for this form, I did find a bug where the drawerQuery wasn't getting passed to the drawer helper. This change, [ButtonControlGroup](https://github.com/NGLPteam/wdp-admin-ui/pull/157/files#diff-c1de5c6e46571289a39fa05e934338e8d354b3b220e8f2c4704cc4a47ddb89beR27), and [DrawerLink](https://github.com/NGLPteam/wdp-admin-ui/pull/157/files#diff-14a158117345830464bb9185b4a81f585fb29f9932272d3ec15202aff399d47aR20) changes fix that bug.

components/composed/contribution/CollectionContributionCreateDrawer/CollectionContributionCreateDrawer.tsx
```diff
@@ -0,0 +1,58 @@
...
+export default function CollectionContributionCreateDrawer({
+  dialog,
+  params,
+}: {
+  dialog: DialogProps;
+  params: Record<string, string>;
+}) {
+  const { t } = useTranslation();
+  const drawerHelper = useDrawerHelper();
+  const slug = useRouteSlug();
```
jen-castiron, line 21: Note: Need to check that we're on the contribution route, or any slug will get passed in to the drawer.

zdavis: I don't think this drawer (or any drawer should be using the route slug), because drawers should be completely separate from the context trigger them. That is, an "add contribution" drawer should work from anywhere, not just from this specific route. So, rather than getting the collection slug from the route, the button that triggers the drawer should include it in the drawerQuery so the drawer can pick it up, similar to what it's doing on edit views.

PR #165, opened 2021-09-28
[B] Fix multiselect dropdown flash

jen-castiron:


components/forms/Multiselect/Multiselect.styles.ts
```diff
@@ -47,11 +49,18 @@ export const List = styled.ul<ListProps>`
...
   overflow-y: auto;
   transform: translateY(${pxToRem("41px")});
-  z-index: var(--z-index-dropdown);
 
-  ${noInsetSupport(`left: 0; right: 0; top: 0;`)}
+  display: none;
+  pointer-events: none;
+  z-index: -1;
 
-  ${({ isOpen }) => !isOpen && `opacity: 0;`}
```
jen-castiron, line 54: This was the culprit - passing in a property to the styled component cause the component to re-render unexpectedly. Once I switched to checking for attributes (`&[open]`), the issue was fixed.

components/forms/Multiselect/Multiselect.styles.ts
```diff
@@ -66,7 +75,9 @@ export const ListItem = styled.li<ListItemProps>`
   padding-block-end: ${basePadding(3)};
   ${tTruncate}
 
-  ${({ highlighted }) => highlighted && `background-color: var(--brand20)`}
```
jen-castiron, line 69: Same here - if you hovered over a list item, the list would flash again. Switching to css attribute selector fixed the issue.

PR #168, opened 2021-09-29
[F] Add confirmation modal to delete button in lists and allow buttons to close dropdown

1aurend:
This PR adds a `ConfirmModalBody` component that follows the Simple Modal design and adds a `ButtonControlConfirm` pattern for rendering buttons that open the modal. Currently, only delete actions use the modal and the `ConfirmModalBody` component is hard-coded into the new pattern, so this likely needs further abstraction.


components/atomic/buttons/ButtonControlGroup/ButtonControlGroup.tsx
```diff
@@ -29,6 +31,15 @@ function ButtonControlGroup({
       >
         {children}
       </ButtonControlDrawer>
+    ) : "modalLabel" in props ? (
```
jen-castiron, line 34: Hmm... to make room for other modals, you could add a `modalConfirm` boolean flag, instead of checking for `modalLabel`. Then `modalLabel` and `modalBody` can be reused in other modal button patterns. What do you think, @zdavis ?

components/composed/ConfirmModalBody/ConfirmModalBody.styles.ts
```diff
@@ -0,0 +1,24 @@
...
+    css`
+      flex-direction: column;
+    `,
+    50
+  )}
+`;
+
+export const ConfirmButton = styled(Button)`
+  width: ${pxToRem(197)};
+  max-width: 90%;
```
jen-castiron, line 23: Instead of setting a strict width on the buttons, set these to be `flex: 1 1 auto`. Then the buttons will fill their container, and the gap can be adjusted as needed! You could even take away the media query and set `flex: wrap;` on the parent, and let the browser decide when the buttons need to stack.

components/composed/ConfirmModalBody/ConfirmModalBody.tsx
```diff
@@ -0,0 +1,37 @@
...
+  handleClose,
+  onClick: onConfirm,
+  actionLabel,
+}: ConfirmModalBodyProps) => {
+  return (
+    <>
+      <p className="t-copy-sm">{body}</p>
+      <Styled.ButtonWrapper>
+        <Styled.ConfirmButton onClick={onConfirm}>
+          {actionLabel ?? "Confirm"}
```
jen-castiron, line 19: These buttons need i18n keys (Apologies if this was already on your list of todos 😅 )

jen-castiron (changes requested):
👏 This is looking great so far! I made a few styling and abstraction suggestions, but I think you did a great job breaking this down into reusable patterns. There is one odd bug I found on mobile:
Screen Shot 2021-09-29 at 9 13 31 AM

1aurend:
> 👏 This is looking great so far! I made a few styling and abstraction suggestions, but I think you did a great job breaking this down into reusable patterns. There is one odd bug I found on mobile: Screen Shot 2021-09-29 at 9 13 31 AM

components/atomic/dropdown/Dropdown/BaseDropdown.tsx
```diff
@@ -29,6 +29,26 @@ function BaseDropdown({ className, disclosure, submenu, label }: Props) {
...
+  const submenuItems = submenu.props.children;
+  const submenuItemsWithModal = submenuItems.map((item: JSX.Element) => {
+    const buttonControl = item.props.children;
+    const buttonControlWithModal = React.cloneElement(
+      buttonControl,
+      onOpenModal
+    );
+    return React.cloneElement(item, {}, buttonControlWithModal);
+  });
+
```
1aurend, line 51: I think we can avoid this mess by having `BaseDropdown` use `forwardRef` and defining `submenu` here rather than up in `Dropdown`. I just did this as POC to make sure I could get the button controls to both open the modal and close this.

components/composed/model/ModelList/hooks/useRowActions.tsx
```diff
@@ -48,14 +53,25 @@ function getButtonProps<D extends Record<string, unknown>>(
...
+        modalBody: actionDefinition.modalBody ?? null,
+      }
+    : {
+        "aria-label": actionDefinition.label,
+        icon: actionDefinition.icon,
+        iconRotate: actionDefinition.iconRotate || 0,
+        ...(actionConfig?.handleClick && {
+          onClick: () => actionConfig.handleClick({ row }),
+        }),
+      };
```
1aurend, line 74: @jen-castiron We should look at this in conjunction with the issue of simplifying `ButtonControl`. It would probably make sense to just pass a `drawer` or `modal` flag as part of the action in `useRowActions` and then instead of making these prop objects, just pass the correct `ButtonControlGroup` child here like you propose doing for button controls in the list headers. I've modeled the pattern for the flag in the action for modal, but we'd have to also implement it for drawer.

components/composed/ConfirmModalBody/ConfirmModalBody.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const ConfirmModalBody = ({
+  modalBody: body,
+  handleClose,
+  onClick: onConfirm,
+  actionLabel,
+}: ConfirmModalBodyProps) => {
+  return (
+    <>
+      {body}
```
1aurend, line 17: This component should be renamed to just `ConfirmModal` but it now takes any JSX as a body prop so actions can pass in their own modal content.

components/atomic/buttons/ButtonControl/patterns/ButtonControlConfirm.tsx
```diff
@@ -0,0 +1,64 @@
...
+  ) => {
+    const dialog = useDialogState({ visible: false, animated: true });
+
+    return (
+      <>
+        <ButtonControl
+          as="a"
+          icon={icon}
+          ref={ref}
+          onClick={(e) => {
```
1aurend, line 34: This took me too long to figure out, but I learned that you can't use Reakit's `<DialogueDisclosure>` and a click handler on the same element, so I had to create my own handler here to both close the dropdown and open the modal. I think the aria properties are all preserved, but I'd love your eye on this @jen-castiron!

components/composed/model/ModelList/hooks/useModelList.tsx
```diff
@@ -68,7 +68,7 @@ function useModelList<
     () => ({
       ...(actions.handleEdit && { edit: { handleClick: actions.handleEdit } }),
       ...(actions.handleDelete && {
-        delete: { handleClick: actions.handleDelete },
+        delete: { handleClick: actions.handleDelete, modalConfirm: true },
```
1aurend, line 71: Adding the flag to actions that require a modal. Could also be passed in from the List components themselves.

jen-castiron (changes requested):
I've made a few comments on your solution, thank you for your work! I was starting to worry that maybe we should just use a different z-index, but when I tried that, the keyboard navigation was broken. So either way, we need to close the dropdown.

I'm still investigating if there's a better way to close the dropdown... Technically, we would want it to close on any button click! If I can't find a solution tonight, I'll try again tomorrow morning.

components/atomic/dropdown/Dropdown/BaseDropdown.tsx
```diff
@@ -29,6 +29,26 @@ function BaseDropdown({ className, disclosure, submenu, label }: Props) {
     hidden: !active,
   };
 
+  const onOpenModal = {
+    onOpenModal: useCallback(
```
jen-castiron, line 33: It took me a minute to get that this is the buttonControlWithModal props. 😅 I'm seeing an error in the console:
```
Warning: Unknown event handler property `onOpenModal`. It will be ignored.
```
This is because the property `onOpenModal` is getting passed directly to the child element, and `onOpenModal` isn't a valid attribute on a button. It still works, just has an annoying error. Maybe it could be avoided with your refactor comment below.

(more info: https://reactjs.org/warnings/unknown-prop.html)

jen-castiron: @1aurend I think I figured it out! Instead of passing `onOpenModal` to the buttons below, why not set an `onClick` event to all the buttons that close the dropdown? Technically, we want the dropdown to close no matter which button is clicked. Since I'm not preventing the default click event from bubbling up, the modal opens as expected.
```
  const onClick = useCallback(() => {
    setActive(false);
  }, [setActive]);

  const submenuProps = {
    id: uid,
    "aria-hidden": !active,
    hidden: !active,
    onClick,
  };
```
You won't need to clone the elements again!

1aurend: @jen-castiron Thanks so much for all of this! And 🎊 for figuring out the custom click handler in `DialogueDisclosure`! I'm tempted to try adding the forward ref to `BaseDropDown` and to continue to pass the on `onClick` to each button in the list individually, probably changing the prop name to something like `closeDropdown` so it's clear what function it is. Then in `ButtonControl` we just always check for that prop and make sure to add it to the `onClick` if it's there? I think if add the forward ref and we're not cloning so many things, it will be a lot clearer what's going on in the dropdown and not so bad having to pass the close prop around to buttons?

1aurend: I'll also look into the unknown event handler issue. I think you're right that the refactor (and renaming it) will resolve that.

components/atomic/buttons/ButtonControl/ButtonControl.tsx
```diff
@@ -7,10 +7,19 @@ type AuthorizeProps = React.ComponentProps<typeof Authorize>;
...
+        ? (e: React.MouseEvent) => {
+            closeDropdown();
+            onClick(e);
+          }
+        : onClick
+      : closeDropdown;
 
     const content = (
-      <Styled.ButtonControl ref={ref} {...props}>
+      <Styled.ButtonControl ref={ref} onClick={handleClick} {...props}>
```
1aurend, line 22: Here I added a general `handleClick` so `ButtonControl` can manage the new optional `closeDropdown` prop.

components/atomic/buttons/ButtonControl/patterns/ButtonControlConfirm.tsx
```diff
@@ -19,28 +19,28 @@ const ButtonControlConfirm = forwardRef(
...
-            if (closeDropdown) {
-              closeDropdown(e);
-            }
-            dialog.show();
-          }}
-        >
-          {children}
-        </ButtonControl>
-        <Modal label={modalLabel} dialog={dialog}>
+        <DialogDisclosure {...dialog} onClick={handleClick}>
```
1aurend: Updated the `handleClick` and reinserted `DialogDisclosure`. Hopefully this fixes the tab issue!

components/atomic/buttons/ButtonControl/patterns/ButtonControlDrawer.tsx
```diff
@@ -7,10 +7,15 @@ type DrawerLinkProps = React.ComponentProps<typeof DrawerLink>;
  * A button control that opens a drawer
  */
 const ButtonControlDrawer = forwardRef(
-  ({ drawer, drawerQuery, icon, children }: Props, ref) => {
+  ({ drawer, drawerQuery, icon, children, closeDropdown }: Props, ref) => {
```
1aurend, line 10: Passing `closeDropdown` through each type of `ButtonControl` pattern

components/atomic/dropdown/Dropdown/BaseDropdown.tsx
```diff
@@ -1,75 +1,81 @@
...
 /**
  * A dropdown for navigation submenus.
  * Adds appropriate aria labels and open/close functionality.
  */
-function BaseDropdown({ className, disclosure, submenu, label }: Props) {
-  const uid = useUID();
-  const wrapperRef = useRef(null);
-  const [active, setActive] = useState(false);
+const BaseDropdown = forwardRef<HTMLUListElement, Props>(
+  ({ className, disclosure, menuItems, out, label }, elRef) => {
```
1aurend, line 12: I learned how to do a `forwardRef` in typescript! ✨

components/atomic/dropdown/Dropdown/BaseDropdown.tsx
```diff
@@ -1,75 +1,81 @@
...
+        {menuItems.map((item, i) => {
+          if (item === null) return null;
+          return (
+            <Styled.Item key={i}>
+              {React.cloneElement(item, { closeDropdown: closeDropdown })}
+            </Styled.Item>
+          );
+        })}
+      </Styled.List>
+    );
```
1aurend, line 39: Here's the new `submenu` with the `forwardRef` and a lot less cloning. I think it's a lot easier to follow now. Since we're passing `closeDropdown` to a `ButtonControlGroup` here, each button in the list can get that prop and thus close the dropdown.

A question, though— now that `submenu` is here, styles are too. So, if we go this route, it's not clear to me whether it's still useful to have two components, `Dropdown` and `BaseDropdown`. Potentially, we could also pull `submenu` out as its own component?

components/atomic/dropdown/Dropdown/Dropdown.tsx
```diff
@@ -10,22 +10,15 @@ import * as Styled from "./Dropdown.styles";
...
+        ref={elRef}
+        out={out}
         className=""
         label={label}
         disclosure={disclosure}
-        submenu={submenu}
+        menuItems={menuItems}
       />
     </Styled.Wrapper>
   );
```
1aurend, line 24: Here's `Dropdown` no longer doing much...

1aurend:
@jen-castiron I went ahead and implemented what I said I'd try above for you to look at. I refactored `BaseDropdown` so that we could create the submenu in there without cloning lots of times, and then I cleaned up passing the `closeDropdown` function to `ButtonControl`. It seemed like the best way to do that was just to have the base `ButtonControl` responsible for handling that prop, so that any button can close a dropdown. My notes above have more detail. While I loved the idea of just putting the click handler on the list, I think this way of doing it, while more complex, is maybe more correct for a11y? Let me know what you think!

jen-castiron (commented):
This is looking so much better! Your comments really helped me understand all of the changes that were made, so thank you for that. I agree that the old `Dropdown` component can be removed. Let's rename `BaseDropdown` to just `Dropdown` and delete the old component.

There's a minor styling issue that should be fixed before merge. The focus state and background color of the delete button should match the Edit button:
Screen Shot 2021-10-01 at 10 10 39 AM

Screen Shot 2021-10-01 at 10 07 24 AM

components/atomic/buttons/ButtonControl/ButtonControl.styles.ts
```diff
@@ -78,3 +79,8 @@ export const ButtonText = styled.span<Pick<BaseProps, "size" | "icon">>`
       padding-inline-end: ${pxToRem("10px")};
     `}
 `;
+
+export const Disclosure = styled(DialogDisclosure)`
+  --dropdown-list-item-padding: ${basePadding(2)}, ${basePadding(6)};
+  --button-background: none;
```
jen-castiron, line 85: @1aurend This line is what's removing the button background from the delete button, and looks like it can be removed 👍

components/atomic/buttons/ButtonControl/patterns/ButtonControlConfirm.tsx
```diff
@@ -0,0 +1,64 @@
...
+      if (closeDropdown) {
+        closeDropdown();
+      }
+      dialog.show();
+    };
+
+    return (
+      <>
+        <Styled.Disclosure {...dialog} onClick={handleClick}>
+          <ButtonControl as="a" icon={icon} ref={ref}>
```
jen-castiron, line 40: @1aurend This will render an a tag wrapped in a button - since we're using the ButtonControl as just styles (and not an a tag), try `as="span"` and see if it behaves correctly.

components/atomic/buttons/ButtonControl/patterns/ButtonControlConfirm.tsx
```diff
@@ -0,0 +1,64 @@
...
+      e.preventDefault();
+      if (closeDropdown) {
+        closeDropdown();
+      }
+      dialog.show();
+    };
+
+    return (
+      <>
+        <Styled.Disclosure {...dialog} onClick={handleClick}>
```
jen-castiron, line 39: @1aurend  This is the component that needs the focus styling applied!

1aurend:
@jen-castiron Thank you for these notes! I got everything except a 2px difference in height between the edit and delete buttons on desktop that I can't figure out. I'm also wondering if there's a way to avoid passing `breakpoint` to `Disclosure`, but I couldn't figure out a way to get the padding right without it. I'll put a note on that in the code so it's easy to find.

components/atomic/buttons/ButtonControl/ButtonControl.styles.ts
```diff
@@ -80,7 +84,31 @@ export const ButtonText = styled.span<Pick<BaseProps, "size" | "icon">>`
...
+      css`
+        --dropdown-list-item-padding: 0;
+
+        > * {
+          padding: ${basePadding(2)} ${basePadding(6)};
+          background: transparent;
+        }
+      `,
+      breakpoint
+    )}
```
1aurend, line 113: It seemed to me that when `Disclosure` inherits a padding from `Dropdown` it needs to ignore it itself and pass it down to its child that's using the main `ButtonControl` styles. But maybe there's a better way?

components/atomic/buttons/ButtonControl/patterns/ButtonControlConfirm.tsx
```diff
@@ -56,9 +62,7 @@ const ButtonControlConfirm = forwardRef(
   }
 );
 
-interface Props extends BaseProps {
-  modalLabel?: ModalProps["label"];
-  modalBody?: React.ReactNode;
-}
+type Props = Pick<GroupProps, "breakpoint"> &
+  BaseProps & { modalLabel?: ModalProps["label"]; modalBody?: React.ReactNode };
```
1aurend, line 66: Is this type okay?

jen-castiron (approved):
@1aurend I see what's going on with the 2px button height issue - the border that's needed for the focus state is transparent, making the button inside look a little smaller. I think there's a better way to share the button styles, but I'll need to play around with it to find the best solution. For now, let's get this squashed merged in!

PR #169, opened 2021-09-29
[F] Add LoadingCircle to edit drawers

1aurend:
Adds a loading state to drawers


components/layout/Drawer/Drawer.tsx
```diff
@@ -20,6 +22,7 @@ const Drawer = ({
   const uidLabel = useUID();
   const uidDesc = useUID();
   const { t } = useTranslation();
+  const loading = useQueryStateContext().loading;
```
jen-castiron, line 25: This is a little bit cleaner way of saying the same thing:
`const { loading } = useQueryStateContext();`

jen-castiron (approved):
Nice work!

PR #170, opened 2021-09-30
[POC] Translation cleanup

jen-castiron:
The problem:
We're repeating ourselves a lot in our translation files, specifically with glossary terms (Contributor, Collection, etc) and field values. This POC cleans up the field values, and shows how we can clean up translations by passing in the glossary term we need, rather than repeating ourselves in the translation document. This could mean easier translations (less words to translate), but a slightly more complex translation lookup for developers. 

For example:
Instead of repeating ourselves in the translation file, we use one "action.edit" string, and pass in the name:
https://github.com/NGLPteam/wdp-admin-ui/pull/170/files#diff-c9218bed23f0aeb93d7e1a7d5f6e1345fae20f4b8692b27fb97fd0f37b3d26d7L43-R44
```
"actions": {
    "edit": "Edit $t({{ name }})"
}
```
Then, where we need to use the string, we pass in the name:
```
<Drawer label={t("actions.edit", { name: "glossary.collection" })} ... />
```
If we use this pattern throughout the translation file, then we only need to translate "Collection" once.

### DIFF
--- a/components/atomic/navs/MobileSubNav/MobileSubNav.stories.tsx
+++ b/components/atomic/navs/MobileSubNav/MobileSubNav.stories.tsx
@@ -43,11 +43,11 @@ const LINKS = [
 const TABS = [
   {
     route: "collection.child.collections",
-    label: "glossary.collection.label_plural",
+    label: "glossary.collection_plural",
   },
   {
     route: "collection.child.items",
-    label: "glossary.item.label_plural",
+    label: "glossary.item_plural",
   },
   {
     route: "collection.manage",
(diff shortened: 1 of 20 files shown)


zdavis:
Is there a best practice here? How do other projects handle this? Will all languages work the same as English, and take the same form of the noun in all cases? Or are there language specific edge cases that could make this approach not hold up? I tend to want to make more work for the translator, and less work for the developer, but I'm really not familiar with best practices around this.

jen-castiron:
I'll have to do a little more research here on best practices and edge cases. You may be right - it might be better to have all cases in translations, just in case.

PR #172, opened 2021-10-01
[B] Remove name label from grids

1aurend:
@jen-castiron This was literally a one-letter fix, but I wanted to flag that problem was matching a case sensitive string in case we want to do more to make these consistent. "Name" seems to be the only `id` value that's capitalized.

### DIFF
--- a/components/composed/model/ModelGrid/ModelGridItem.tsx
+++ b/components/composed/model/ModelGrid/ModelGridItem.tsx
@@ -30,7 +30,7 @@ function ModelGridItem<T extends Record<string, unknown>>({ row }: Props<T>) {
 
           return (
             <div key={i}>
-              {cell.column.id !== "title" && cell.column.id !== "name" && (
+              {cell.column.id !== "title" && cell.column.id !== "Name" && (
                 <span>
                   {cell.render("Header")}
                   {`: `}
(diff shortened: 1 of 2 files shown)


jen-castiron:
That is odd! Here's where the "Name" ID is located: https://github.com/NGLPteam/wdp-admin-ui/blob/main/components/composed/model/ModelColumns/NameColumn.tsx#L17
I don't see why it's capitalized, and should probably be set to lower case. To avoid any further issues, we could take the column.id string, set it to lower case, and then check against "title" and "name"

components/composed/model/ModelGrid/ModelGridItem.tsx
```diff
@@ -30,12 +30,13 @@ function ModelGridItem<T extends Record<string, unknown>>({ row }: Props<T>) {
...
           return (
             <div key={i}>
-              {cell.column.id !== "title" && cell.column.id !== "name" && (
-                <span>
-                  {cell.render("Header")}
-                  {`: `}
-                </span>
-              )}
+              {cell.column.id.toLowerCase() !== "title" &&
+                cell.column.id !== "name" && (
```
jen-castiron, line 34: Need to check against `.toLowerCase` for "name" as well, and then you're all set!

PR #173, opened 2021-10-01
[WIP][F] List users with roles on collections and items

jen-castiron:
- Resolves WDP-222


hooks/useDestroyer.ts
```diff
@@ -238,18 +253,42 @@ const destroyOrderingMutation = graphql`
...
   }
 `;
 
 const destroyFragment = graphql`
-  fragment useDestroyerFragment on DestroyMutationPayload @inline {
-    destroyed # can be boolean or nil, basically just a truthy check
-    globalErrors {
-      message
-      type
+  fragment useDestroyerFragment on StandardMutationPayload @inline {
```
jen-castiron, line 271: @1aurend I figured out how to use the same destroyer fragment for disable and revoke! You specify which kind of payload you expect, and their properties. Now that they're all declared here in one fragment, we can clean up our types at the top here: https://github.com/NGLPteam/wdp-admin-ui/pull/173/files#diff-adedaedf553bc89fd26d37d2753c78a8a5c1c02b422128335e64bfeba3ac03a6L41-R40

PR #174, opened 2021-10-01
[F] Improve user experience around editing details & schema properties

jen-castiron:
- Adjust form grid two column layout
 - Add optional cancel buttons to forms
 - Add schema form to Collection & Item detail view
 - Add cancel button to Collection & Item update drawer
 - Remove Schema from sidebar menu
 - Add header buttons to collection drawer
 - Add "Save & Close" button
 - Add Schema form to Collection & Item update drawer


jen-castiron:
For the drawer actions, I created this component that can be reused with different models:
https://github.com/NGLPteam/wdp-admin-ui/pull/174/files#diff-f97d3bcdbdf2e1d307110bf7e3fb004d5b841b941e0688d7c1f9fe9d639800c4
The idea is that this component can be dropped in, given a few routes and optional delete action, and it will create a row of button. 

Right now I'm passing the model's child routes, but that doesn't quite match up with what's in the designs. 
App:
Screen Shot 2021-10-04 at 2 49 27 PM

Designs:
Screen Shot 2021-10-04 at 2 57 43 PM

The designs have a "Detail View", "View Collections", and "View Items". This would mean we'd have to hard code some routes in the drawer to match designs. Something like:
```
const routes = [{
   route: 'collection.manage.details',
   label: 'navLabels.details',
   query: { slug: drawerSlug }
}, {
   route: 'collection.child.collections',
   label: 'glossary.collections.view_label',
   query: { slug: drawerSlug }
}, 
... ];
```
Alternatively, I could define the drawer routes inside `baseRoutes`, but that didn't seem like the right place. I could be wrong though! @zdavis, I know you want to avoid hard coding routes, so let me know what you think.

PR #175, opened 2021-10-04
[WIP] Wire up delete buttons for orderings in list

1aurend:


hooks/useDestroyer.ts
```diff
@@ -137,13 +146,33 @@ export function useDestroyer() {
...
+  const ordering = useCallback(
+    async (input: DestroyOrderingInput, label: string) => {
+      const response = await commitDisableOrDestroyOrdering({
+        variables: { input },
+      });
+      return handleResponse(
+        response.destroyOrdering,
+        label,
+        response.destroyOrdering?.disabled
+      );
```
1aurend, line 163: @jen-castiron I'm struggling with the types in `handleResponse` so that I don't have to pass `disabled` in separately like this, since `disabled` doesn't exist on `useDestroyerFragment`. What's a better way to handle this? I'm also happy for any other feedback you have on this when you get a chance. Thanks!

hooks/useDestroyer.ts
```diff
@@ -23,22 +23,30 @@ import type {
...
+    (data: useDestroyerFragment$key | DestroyOrdering | null, name: string) => {
       if (!data) return;
       const results = readInlineData<useDestroyerFragment>(
         destroyFragment,
         data
       );
-      if (results.destroyed) {
+      if ("disabled" in data && data.disabled) {
+        notify.success(t("outcomes.disabled", { name }));
+      } else if (results.destroyed) {
```
1aurend, line 49: Updated types to this

hooks/useDestroyer.ts
1aurend: I spent some time working on the types and came up with something I think is better—see comment below.

locales/en.json
```diff
@@ -65,7 +65,8 @@
       }
     },
     "outcomes": {
-      "deleted": "$t({{ name }}) deleted successfully."
+      "deleted": "$t({{ name }}) deleted successfully.",
+      "disabled": "$t(Default ordering {{ name }}) disabled successfully."
```
jen-castiron, line 69: easy fix - $t() translates whatever it wraps, so it should only wrap `{{ name }}`

hooks/useDestroyer.ts
jen-castiron: 👀 This looks good to me, but @scryptmouse might have another solution. Can `disabled` be added to the `useDestroyerFragment`? @scryptmouse , will be other mutations where `disabled` is returned instead of `destroyed`?

jen-castiron (commented):
I'm curious what Alexa thinks about the types, and if `disabled` can be added to the general destroy fragment. I suspect there will be other mutations that return disabled instead of destroyed. 

While testing, I found that there's no message returned if you disable an ordering twice. The first time I disabled[ this "Everything" order](https://wdp-admin-client-j8fgd9pq4-cast-iron.vercel.app/collections/E437895cry5Zia6XcG7oXF2NdPvo5nO/manage/order), I saw the "disabled successfully" message. If you try to disable it again, the API returns a 200 and `disabled: false, destroyed: false`. @scryptmouse , is this expected API behavior? We might need to either disable the delete button if an ordering is already disabled, and/or give the ordering a disabled style.

PR #179, opened 2021-10-07
[F] Make it possible to set/update an entity's schema

jen-castiron:
Edit a collection or item to change the schema. If needed, we can hide the schema selector from drawers. 

Note: Changing the schema is currently a destructive action!


pages/items/[slug]/manage/details.tsx
```diff
@@ -39,6 +42,10 @@ const query = graphql`
       ...ItemLayoutFragment
       ...ItemUpdateFormFragment
       ...SchemaInstanceFormFragment
+      ...SchemaSelectorDataFragment
+    }
+    schemaVersions {
+      ...SchemaSelectorOptionsFragment
```
Comment (jen-castiron) [inline pages/items/[slug]/manage/details.tsx:48]: I've taken a slightly different tact here with the schema selector dropdown. I added the query for the schema versions in the page / drawer queries.
Pros to this pattern:
1. Adding fragments is relatively easy
2. Adding new properties to parent wrappers is relatively easy
3. This input data is loaded with the page query, and doesn't need an extra query to render

In the other typeahead inputs (Collections, Items, etc) the query wraps the input. ([See Collection Typeahead code](https://github.com/NGLPteam/wdp-admin-ui/blob/main/components/forms/CollectionTypeahead/CollectionTypeahead.tsx#L34-L45))
Pros to this pattern are:
1. The input can be added anywhere, without shuffling fragments around
2. If (for whatever reason) the query for the list fails, the entire page or drawer query doesn't fail
3. Less properties need to be passed all the way down from the parent to the input

@zdavis & @scryptmouse - I'm curious what your thoughts are on these two approaches, and if you have a preference.

PR #180, opened 2021-10-07
[B] Fix a11y issues

jen-castiron:
- Add missing alt text in thumbnail images
- Add missing delete button text
- Add missing lang attribute

### DIFF
--- a/components/atomic/buttons/ButtonControl/patterns/ButtonControlConfirm.tsx
+++ b/components/atomic/buttons/ButtonControl/patterns/ButtonControlConfirm.tsx
@@ -1,9 +1,8 @@
 import React, { forwardRef } from "react";
 import { ButtonControl, ButtonControlGroup } from "components/atomic";
-import { useDialogState } from "reakit/Dialog";
+import { DialogDisclosure, useDialogState } from "reakit/Dialog";
 import Modal from "components/layout/Modal";
 import ConfirmModal from "components/composed/ConfirmModal";
-import * as Styled from "../ButtonControl.styles";
 
 type BaseProps = React.ComponentProps<typeof ButtonControl>;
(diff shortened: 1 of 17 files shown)


components/atomic/buttons/ButtonControl/patterns/ButtonControlConfirm.tsx
```diff
@@ -38,15 +36,16 @@ const ButtonControlConfirm = forwardRef(
 
     return (
       <>
-        <Styled.Disclosure
+        <DialogDisclosure
+          as={ButtonControl}
```
jen-castiron, line 40: @1aurend Something I totally forgot about Reakit components - they can be styled by setting `as` prop to a styled component! I updated the delete button to use this pattern, so we don't have to repeat styles. ✨ I didn't see any mobile issues, but let me know if something accidentally regressed.

1aurend: Nice @jen-castiron! It looks great to me!

1aurend: Follow up— I'm working on implementing the `ButtonControlGroup` that accepts children and I've come across two related things:
- `ButtonControlDrawer` appears not to be tabbable?
- the styles in `ButtonControlDrawer` need to be modified in parallel to what you've done here for `ButtonControlConfirm` so the icons line up in a dropdown:

Screen Shot 2021-10-08 at 10 25 12 AM

At this point, the only place we'd be using `ButtonControlDrawer` in a dropdown is storybook so it's not urgent.

1aurend: I'm using Firefox

PR #182, opened 2021-10-08
[FIXUP] Alternate approach to schema version options fragment

zdavis:


pages/items/[slug]/manage/details.tsx
```diff
@@ -44,8 +44,6 @@ const query = graphql`
       ...SchemaInstanceFormFragment
       ...SchemaSelectorDataFragment
     }
-    schemaVersions {
-      ...SchemaSelectorOptionsFragment
-    }
+    ...SchemaInstanceFormSchemaOptionsFragment
```
Comment (jen-castiron) [inline pages/items/[slug]/manage/details.tsx:47]: Nice! less to remember on parent queries. I can add this to my PR and get it built early Monday, before demo.

Comment (zdavis) [reply pages/items/[slug]/manage/details.tsx:47]: Sweet! Glad it helps! I only made this change for the one detail view — I imagine it will have to be applied to drawers, collection, etc.

PR #189, opened 2021-10-15
[F] Wire up Add/Delete Link buttons for items and collections

1aurend:
@jen-castiron I've added the "Add Link" button and wired up a basic version of the relevant drawer for the Links tab on both Items and Collections. You can't yet delete the links, but that should be straightforward to implement once the add form is finalized. I could use your input on the form structure and labels, however:
- I wasn't sure if the title of the drawer should be the source item/collection or just "Add Link". 
- It seems like the user might need some further clue about how to understand the "Link Type" choices. I assume it's supposed to be `[sourceEntity][operator][targetEntity] ` but do we need to make that clearer in the form? 
- In addition, the UX has a single input for "Collection or Item", but it looks like there are only separate typeaheads for these two? Let me know if I missed something. I'm using the `ItemTypeahead` as a placeholder for now.


scryptmouse:
I can definitely add a new endpoint specifically for typeaheads like that, but I think it might be nice to have it be something more specific / applicable. What if it is an endpoint that works like:

```graphql
linkTargets(sourceId: ID!, needle: String) {
  ... on Collection {
    id
    title
  }

  ... on Item {
    id
    title
  }
}
```

The benefit of this approach is I could make sure that you don't get anything that would make an invalid link for the provided source, e.g. none of its immediate children (which don't make sense to link), which would reduce any filtering needs on the frontend.

Rather than a top-level field, I could also make it a property on any existing entity if you'd prefer not to have to manage passing the entity's id to the field.

1aurend:
Thanks @scryptmouse @jen-castiron! It makes sense to me to have this new endpoint be a property of `Entity` since we already have to query to get the `sourceEntity` id when we open the drawer, we may as well add a fragment to get its valid link targets then?

components/forms/LinkTargetTypeahead/LinkTargetTypeahead.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+type LinkTargetEdge = LinkTargetTypeaheadFragment$data["edges"][number];
+
+export default LinkTargetTypeahead;
+
+const fragment = graphql`
+  fragment LinkTargetTypeaheadFragment on LinkTargetCandidateConnection {
+    edges {
+      node {
+        kind
```
jen-castiron, line 70: nit: Looks like `kind` can be removed since it's not used in this component

jen-castiron (commented):
This is looking good! Just added a couple nit comments.

PR #193, opened 2021-10-19
[F] Clicking thumbnail opens relevant entity page

1aurend:
Adds links to thumbnail images on list pages

### DIFF
--- a/components/composed/model/ModelColumns/ThumbnailColumn.tsx
+++ b/components/composed/model/ModelColumns/ThumbnailColumn.tsx
@@ -31,7 +31,7 @@ const ThumbnailColumn = <NodeType extends Node>({
   return {
     Header: <span className="a-hidden">{t("columns.thumbnail")}</span>,
     id: "thumbnail",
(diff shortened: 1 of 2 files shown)


components/composed/model/ModelColumns/ThumbnailColumn.tsx
```diff
@@ -17,28 +17,45 @@ export interface Thumbnail {
...
+  route,
+  ...props
+}: ThumbnailColumn<NodeType>): Column<NodeType> => {
   const { t } = useTranslation();
 
   return {
     Header: <span className="a-hidden">{t("columns.thumbnail")}</span>,
     id: "thumbnail",
-    accessor: (row: NodeType) => row?.thumbnail?.image,
+    accessor: (thumbnail: NodeType) => thumbnail?.image,
```
jen-castiron, line 34: These lines are breaking the image - if you change it back to `(row: NodeType) => row?.thumbnail?.image,`, the images will appear again. The accessor passes in a node (`row`) and we need to get the thumbnail object from that row.

PR #196, opened 2021-10-20
[F] Redirect to items if collection does not have child collections

jen-castiron:
Query for a collection's children before redirecting to child collections. If no child collections are found, redirect to child items.
- Resolves WDP-289

### DIFF
--- a/pages/collections/[slug]/index.tsx
+++ b/pages/collections/[slug]/index.tsx
@@ -1,8 +1,57 @@
-import { Page } from "types/page";
+import { graphql } from "react-relay";
+import { useRouteSlug } from "hooks";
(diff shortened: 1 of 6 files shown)


pages/collections/[slug]/index.tsx
```diff
@@ -1,8 +1,57 @@
...
+      query: { slug },
+    });
+  } else if (total === 0 && itemRoute) {
+    router.push({
+      pathname: itemRoute.path,
+      query: { slug },
+    });
+  }
+
+  return <LoadingCircle />;
```
Comment (jen-castiron) [inline pages/collections/[slug]/index.tsx:35]: Note to self: Need to account for error states

PR #206, opened 2021-10-29
[F] Build Add Collection form

1aurend:
Adds a form adding a new collection to a community or collection entity and implements the relevant button on those entity pages.

**Should maybe wait to merge after #202 is merged, so we can double check the order of the buttons in the collection entity page header


jen-castiron (approved):
Looking good! Can you add a delete button, like you did for items? Doesn't have to be in this commit, but I thought it was really smart to add that in at the same time as "Add Item".

1aurend:
@jen-castiron I rebased after merging #202 and added the delete button, but there's an issue with the modal not closing that didn't come up on items. If I pass in `dialog.hide` to the click handler (in `ButtonControlConfirm`), the modal closes, but I couldn't figure out how to do that without messing up the type on the click handler for the other modals. Ultimately, though, that didn't seem like the correct solution anyhow. The redirect should close the modal, I think? Would you mind taking a look before I merge?

PR #214, opened 2021-11-02
[F] Add files to collections, add file delete button

jen-castiron:


jen-castiron:
@zdavis for tomorrow - Do you know where the table actions are ordered? I'd like the edit and delete buttons to be the last actions. Currently 'download' is the last action in the files list.

zdavis:
I think the order the actions are defined on the hash in `useRowActions` is what determines the order. see line 38 of `packages/admin/components/composed/model/ModelList/hooks/useRowActions.tsx`. See if defining download first resolves it. I think this file has changed a bit since I wrote it, so this may not be the case any longer.

PR #217, opened 2021-11-03
[F] Build add collection button for collections landing page

jen-castiron:


packages/admin/components/composed/collection/CollectionAddDrawer/CollectionAddDrawer.tsx
```diff
@@ -18,33 +19,57 @@ export default function CollectionAddDrawer({
...
+    }
+    community(slug: $parentSlug) {
+      id
+    }
+    ...CollectionAddFormFragment
+  }
+`;
+
+const rootQuery = graphql`
+  query CollectionAddDrawerRootQuery($schemaKind: SchemaKind!) {
```
jen-castiron, line 72: Both @1aurend  and I felt it was really silly to have to forms for creating a Collection on the root page and interior collection pages, but the implementation is a little funky. The parentSlug is a required field if you are running the collection or community query, so we can't pass an empty value to parentSlug. This is true even if you use the graphql `includes` directive to conditionally fetch this information. 

So what I have here is two different Query wrappers, one with the parentSlug and one without. This means we can use the same CollectionAddForm for both, and only render the Communities list if a parentId is not passed.

PR #220, opened 2021-11-04
[B] Fix pagination styling in Firefox

1aurend:
Adds a new appearance class for pagination input styles.

### DIFF
--- a/packages/admin/theme/utility/appearance.ts
+++ b/packages/admin/theme/utility/appearance.ts
@@ -54,12 +54,6 @@ export default css`
 
   .a-pagination-input {
     max-width: 30px;
-    -moz-appearance: textfield;
-    text-align: left;
-    &:hover,
-    &:focus {
-      -moz-appearance: auto;
-    }
   }
 
   .a-color-accent {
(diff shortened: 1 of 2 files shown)


packages/admin/theme/utility/appearance.ts
```diff
@@ -52,6 +52,16 @@ export default css`
...
+  .a-pagination-input {
+    max-width: 30px;
+    -moz-appearance: textfield;
+    text-align: left;
+    &:hover,
+    &:focus {
+      -moz-appearance: auto;
+    }
+  }
+
```
1aurend, line 64: @jen-castiron Does this look okay? I'm definitely still learning the details of styling forms.

1aurend: Hmm... the linter doesn't like the `-moz` prefixes, but I think they're needed here?

jen-castiron: I don't think the bug here is the numeric controls (which are good for accessibility), just the width. (Check with Zach to be sure.) If the issue is the numeric controls, then you'd want to use just `appearance: textfield` without the mod prefix.

PR #221, opened 2021-11-04
[B] Improve styling of selects in Firefox

1aurend:
Fixed: text alignment

To Do: background color on select


jen-castiron:
Just making a note here - the background styling is different between Firefox and Chrome because the browsers handle focus-visible differently. The state we want is the same - if the user navigates to the select input using their keyboard, the background changes.

PR #224, opened 2021-11-05
[F] Build Add Dynamic Ordering form

1aurend:
First working version of the Add Dynamic Ordering form!

Two known issues:
- `identifier` is not guaranteed to be unique. Need to add a uid.
- `OrderingSelectDefintionInput` only takes the value selected for Collections.

### DIFF
--- a/packages/admin/components/composed/ordering/EntityOrderingAddForm/EntityOrderingAddForm.tsx
+++ b/packages/admin/components/composed/ordering/EntityOrderingAddForm/EntityOrderingAddForm.tsx
@@ -52,10 +52,8 @@ export default function EntityOrderingAddForm({
         .map((node) => node.node.identifier);
   };
 
-  const getOrder = (
-    property: string,
-    direction: string
-  ): OrderDefinitionInput[] => {
+  const getOrder = (sortby: string): OrderDefinitionInput[] => {
+    const [property, direction] = sortby.split(",");
     const directionValue = direction === "asc" ? "ASCENDING" : "DESCENDING";
 
     interface PathOptions extends Record<string, string> {
@@ -88,7 +86,7 @@ export default function EntityOrderingAddForm({
         select: {
           direct: data.collections !== "NONE" ? data.collections : data.items,
         },
-        order: getOrder(data.sortby.split(",")[0], data.sortby.split(",")[1]),
+        order: getOrder(data.sortby),
       },
     }),
     []
@@ -122,8 +120,8 @@ export default function EntityOrderingAddForm({
         />
         <Forms.Fieldset
           label={t("forms.fields.include")}
-          noGap
           description={description}
+          style={{ "--fieldset-legend-padding": 0 }}
         >
           <Forms.RadioGroup
             label="glossary.collection_plural"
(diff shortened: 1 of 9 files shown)


packages/admin/components/forms/BaseSelect/BaseSelect.tsx
```diff
@@ -29,7 +29,7 @@ const BaseSelect = forwardRef(
 
 interface Option {
   readonly label: string;
-  readonly value: string | number;
+  readonly value: string | number | string[];
```
1aurend, line 32: I will remove this before merging.

packages/admin/components/forms/Fieldset/Fieldset.styles.ts
```diff
@@ -11,10 +11,15 @@ export const Fieldset = styled.fieldset`
   flex-basis: var(--form-grid-item-width-wide);
 `;
 
-export const Legend = styled.legend`
-  padding-block-end: var(--fieldset-field-spacing, ${pxToRem(24)});
+export const Legend = styled.legend<LegendProps>`
+  --fieldset-field-spacing: ${({ $noGap }) => ($noGap ? 0 : pxToRem(24))};
```
1aurend, line 15: Okay to set `--fieldset-field-spacing` here?

packages/admin/components/composed/ordering/EntityOrderingAddForm/EntityOrderingAddForm.tsx
```diff
@@ -45,16 +52,43 @@ export default function EntityOrderingAddForm({
...
+          ? data.name.toLowerCase().split(" ").join("")
+          : "",
         filter: {
-          schemas: getFilter(data.directItems, data.directCollections),
+          schemas: getFilter(data.items, data.collections),
         },
+        select: {
+          direct: data.collections !== "NONE" ? data.collections : data.items,
+        },
+        order: getOrder(data.sortby.split(",")[0], data.sortby.split(",")[1]),
```
1aurend, line 91: I don't love this. Is there a better way?

packages/admin/components/forms/Fieldset/Fieldset.styles.ts
jen-castiron: Unfortunately, this would mean that `--fieldset-field-spacing` can't be defined by a parent component, because you're redefining it here in Fieldset. If a parent component needs the gap to be 0, it should set `--fieldset-field-spacing: 0`.

jen-castiron: Side note: I make this mistake a lot too. 😅 I think we should define our variables in `root.ts`, and keep the naming structure the same (`--[component-name]-[property]`). That way we have a central location for all of the CSS variables that are used in the app. @dananjohnson , what do you think?

packages/admin/components/composed/ordering/EntityOrderingAddForm/EntityOrderingAddForm.tsx
jen-castiron: I would pass sortby to getOrder, then do the splitting in the getOrder function:
```
const getOrder = (sortby: string) => {
   [property, direction] = sortby.split(",");
   ...
}
...
order: getOrder(data.sortby)
```
You might need to make some adjustments if order can be null or undefined, but I think that's pretty close.

packages/admin/components/forms/Fieldset/Fieldset.styles.ts
dananjohnson: I'm not sure I understand what you're defining in `root.ts`. You'd have a bunch of `--componentName-property` defined at `:root {}` there? I do like the naming convention. I might even match the casing of the CSS property to the component (i.e. `--Fieldset-field-spacing`) to more clearly indicate the relationship.

PR #228, opened 2021-11-08
[C] Move shared mixins and functions into common/theme

jen-castiron:


packages/admin/package.json
```diff
@@ -37,6 +37,7 @@
     "@uppy/file-input": "^1.5.2",
     "@uppy/react": "^1.12.2",
     "@uppy/tus": "^1.9.2",
+    "@wdp/theme": "1.0.0",
```
jen-castiron, line 40: @zdavis How do you want to name the common packages? Should this be `@wdp/theme`, or `@wdp/common-theme`?

zdavis: I think it should just be @wdp/theme. Less typing is better.

PR #236, opened 2021-11-12
[F] Simplify dynamic ordering form select inputs

1aurend:
Combines `Collections` and `Items` inputs to a single `Entities to Include` input. Creates `identifier` as a slug based on `displayName`. Removes `filter`.

### DIFF
--- a/packages/admin/components/composed/ordering/EntityOrderingAddForm/EntityOrderingAddForm.tsx
+++ b/packages/admin/components/composed/ordering/EntityOrderingAddForm/EntityOrderingAddForm.tsx
@@ -1,19 +1,19 @@
 import * as React from "react";
 import { graphql, useFragment } from "react-relay";
-import { useTranslation } from "react-i18next";
+import { useUID } from "react-uid";
 import MutationForm, {
   useRenderForm,
   useToVariables,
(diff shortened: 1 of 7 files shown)


packages/admin/components/composed/ordering/EntityOrderingAddForm/EntityOrderingAddForm.tsx
```diff
@@ -1,54 +1,53 @@
...
+   *    if (items === "NONE")
+   *      return schemaOptions.edges
+   *        .filter((node) => node.node.kind === "ITEM")
+   *        .map((node) => node.node.identifier);
+   *    if (collections === "NONE")
+   *      return schemaOptions.edges
+   *        .filter((node) => node.node.kind === "COLLECTION")
+   *        .map((node) => node.node.identifier);
+   *  };
+   */
```
1aurend, line 50: @jen-castiron Should I remove everything related to the schema filter before merging?

jen-castiron: Yeah, probably... We don't know if the API will change before we get to it

PR #237, opened 2021-11-12
[C] Add keycloak provider to frontend

jen-castiron:
- Moved shared keycloak code to the `lib` folder
- Setup frontend keycloak provider (same as admin)


jen-castiron:
@zdavis or @scryptmouse ...

This URL flashes when loading the frontend: 
```
http://localhost:3001/#error=login_required&state=f8effb9e-9840-47c9-b994-750e1f648ccf
```

I'm not sure what's different between the two setups, or if I missed a step? if you have a minute, please take a glance and let me know if I'm missing something. Login should not be required, same as the backend.

PR #240, opened 2021-11-15
[F] Build InstallationName component

1aurend:
also includes `CommunityPicker` and `Dropdown` components


packages/frontend/components/atomic/CommunityPicker/CommunityPicker.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import { Story } from "@storybook/react";
+import CommunityPicker from ".";
+type Props = React.ComponentProps<typeof CommunityPicker>;
+
+export default {
```
jen-castiron, line 5: nit: Since the community picker uses the Dropdown component, let's remove the CommunityPicker story.

packages/frontend/components/atomic/Dropdown/Dropdown.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  &:hover,
+  &:focus-within:not(:hover) {
+    .a-bg-brand90 & {
+      background: none;
+    }
+  }
+
+  > * {
+    display: flex;
```
jen-castiron, line 63: nit: This should be a plain block for now, instead of a flex block. The dropdown items don't have icons, so there's no need for flex.

packages/frontend/components/atomic/Dropdown/Dropdown.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+`;
+
+export const List = styled.ul<ListProps>`
+  --dropdown-list-item-padding: ${pxToRem(8)} ${pxToRem(24)};
+
+  position: absolute;
+  top: 0;
+  left: 100%;
+  margin-inline-start: ${pxToRem(8)};
+  display: flex;
```
jen-castiron, line 20: nit: I don't think we need flex here either, since the dropdown items stack on each other.

jen-castiron (commented):
Looking good so far! I've added a few nit comments. 👍

PR #241, opened 2021-11-15
[F] Build CookiesBanner component

1aurend:


packages/frontend/components/global/CookiesBanner/CookiesBanner.styles.ts
```diff
@@ -0,0 +1,16 @@
...
+import { pxToRem } from "@wdp/lib/theme/functions";
+
+export const Banner = styled.div`
+  padding-block: ${pxToRem(36)};
+`;
+
+export const Content = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
```
jen-castiron, line 11: For now, allow this flex area to wrap so the buttons are below when we run out of space. We'll refine this as Lael designs the mobile layouts. 😄

packages/frontend/components/global/CookiesBanner/CookiesBanner.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from "react";
+import * as Styled from "./CookiesBanner.styles";
+
+export default function CookiesBanner({ onAccept, onClose }: Props) {
+  return (
+    <Styled.Banner className="l-container-max a-bg-custom20" role="alert">
```
jen-castiron, line 6: I'd remove the `l-container-max` class here - it's very likely the color will extend edge to edge on the page. I'll double check with Lael 👍

packages/frontend/components/global/CookiesBanner/CookiesBanner.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <Styled.Content className="l-container-wide">
+        <p className="t-copy-sm">
+          Cookies are delicious. Vitae pharetra maecenas amet eu luctus
+          convallis. Feugiat pulvinar eleifend eu fames accumsan, ipsum diam.
+        </p>
+        <Styled.Buttons className="l-flex l-flex--gap">
```
jen-castiron, line 12: The gap here is 12px instead of 16px (applied by `l-flex--gap`) - I'd define the gap in `Styled.Buttons` for now, until we start to see a pattern develop.

jen-castiron:
@1aurend I'm going to merge this in so I have the button --sm classes you defined - can you tackle the small styling issues in another PR?

PR #265, opened 2021-11-30
[F] Style IssueHero

1aurend:


jen-castiron (commented):
This is looking good - same nit comment here about the root padding and flex as the Journal Hero styling.

jen-castiron (approved):
@1aurend Thanks for making those updates! Let's get this merged in - it can always be updated when the API is ready

PR #266, opened 2021-11-30
[F] Style JournalHero

1aurend:


packages/frontend/components/composed/journal/JournalHero/JournalHero.styles.ts
```diff
@@ -1,6 +1,30 @@
...
 `;
+
+export const TitleBlock = styled.div`
+  display: flex;
+  flex-direction: column;
+  max-width: ${pxToRem(670)};
+`;
+
+export const Title = styled.h2`
+  margin-block-end: ${pxToRem(16)};
```
jen-castiron, line 16: @1aurend There are some common padding values that are defined in root: https://github.com/NGLPteam/wdp-client/blob/main/packages/frontend/theme/base/root.ts#L159-L164
```
/* padding */
    --padding-lg: ${pxToRem(32)};
    --padding-md: ${pxToRem(24)};
    --padding-rg: ${pxToRem(16)};
    --padding-sm: ${pxToRem(8)};
    --padding-xs: ${pxToRem(6)};
```
I wish I had made these in admin! Let's use these variables where we can.

packages/frontend/components/composed/journal/JournalHero/JournalHero.styles.ts
```diff
@@ -1,6 +1,30 @@
...
+import { pxToRem } from "@wdp/lib/theme/functions";
 
 export const HeroInner = styled.div`
   padding-block-start: var(--container-padding-lg);
   padding-block-end: var(--container-padding-lg);
 `;
+
+export const TitleBlock = styled.div`
+  display: flex;
+  flex-direction: column;
```
jen-castiron, line 11: Let's not use flex unless we have to - A div aligns it's child items in a column by default. An exception is made if we're modifying the child item's size or alignment using flex.

jen-castiron (commented):
Just a few nits - otherwise the styling looks good!

jen-castiron:
@1aurend When you're ready, merge this in! We can update later when the API is ready.

PR #270, opened 2021-12-02
[F] Build Journal Navigation Bar

1aurend:
Adds `JournalNavBar`. Also moves `Search` into `/forms` and adds a basic route for collection pages.


packages/frontend/components/composed/journal/JournalNavBar/JournalNavBar.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+`;
+
+export const LeftSide = styled.div``;
+
+export const RightSide = styled.div`
+  max-width: ${pxToRem(376)};
```
jen-castiron, line 20: The search bar only grows to the size of it's parent, so right now it's about 250px wide. If you set this to `flex: 1 1 auto`, then this div will grow if there's room, and the search bar will be the right width on desktop 👍

jen-castiron (approved):
Looks good! There's one small styling issue, but once that's fixed it's ready to be squished & merged 🚢

PR #278, opened 2021-12-06
[WIP][F] Build issue ordering dropdown

jen-castiron:
~Ran into a couple questions while working on this:~
1. ~Should we stick to a single query when getting lists of items? In this example, the entire page reloads when you change ordering selection. If we used two queries, then only the item list would need to refresh. I had hoped [refetch container](https://relay.dev/docs/v3.0.0/refetch-container/) would only refetch the components that need the fragment, but it only fetched the fragment data, reloaded the page, and broke the layout. It's entirely possible that it was user error on my part.~
2. ~I'm assuming we always be fetching an ordering on the issue page... What is the default ordering?~
3. ~Should orderings have their own route, or is it OK to use query parameters?~

~Alexa & Zach added as reviewers to discuss above questions~ ☝️ 

All Qs answered in our meeting


packages/frontend/pages/items/[slug]/index.tsx
```diff
@@ -4,16 +4,19 @@ import { QueryWrapper } from "@wdp/lib/api/components";
 import { useRouteSlug } from "@wdp/lib/routes";
 import CommunityChildLayout from "components/composed/community/CommunityChildLayout";
 import { SlugItemQuery as Query } from "@/relay/SlugItemQuery.graphql";
-import EntityLayoutFactory from "components/factories/EntityLayoutFactory";
+import EntityContentLayoutFactory from "components/factories/EntityContentLayoutFactory";
```
Comment (jen-castiron) [inline packages/frontend/pages/items/[slug]/index.tsx:7]: This can probably get renamed to EntityContentIndexFactory or something like that. Not sure if "Content" is clear in this case.

packages/frontend/components/factories/EntityContentLayoutFactory/EntityContentLayoutFactory.tsx
```diff
@@ -0,0 +1,78 @@
...
+    ... on Collection {
+      schemaDefinition {
+        identifier
+      }
+
+      ...EntityLayoutFragment
+      ...JournalLayoutFragment
+      ...JournalInfoFragment
+      ...IssueLayoutFragment
+      ...IssueSummaryFragment
```
jen-castiron, line 66: I really hate adding all these fragments when we only need one, but I need to know the schema to determine which fragment to use.

packages/frontend/components/composed/issue/IssueSummary/IssueSummary.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface Props {
+  data?: IssueSummaryFragment$key | null;
+}
+
+const fragment = graphql`
+  fragment IssueSummaryFragment on Collection {
+    items {
+      ...IssueSummaryListFragment
+    }
+    ordering(identifier: $orderIdentifier) {
```
jen-castiron, line 33: I think it might be safe to assume that we want to always use ordering on this page. In that case, the frontend or the API will need to know what the default ordering is.

PR #303, opened 2021-12-15
[F] Build Prev-Next Issue block

1aurend:
Builds design for Prev-Next Issue block. Query needs to be updated and links added when the API is ready. Also, adds new Button pattern for PrevNextButton.

### DIFF
--- a/packages/frontend/components/composed/issue/PrevNextIssue/PrevNextIssue.styles.ts
+++ b/packages/frontend/components/composed/issue/PrevNextIssue/PrevNextIssue.styles.ts
@@ -1,10 +1,20 @@
 import styled from "styled-components";
 import { pxToRem } from "@wdp/lib/theme/functions";
-import { Button } from "components/atomic";
-import { aButton } from "theme/mixins/appearance";
+import { respond } from "theme/mixins";
 
 export const Inner = styled.div`
-  padding-block: var(--container-padding-lg);
+  padding-block-start: var(--container-padding-lg);
+  padding-block-end: var(--container-padding-lg);
+
+  ${respond(
+    `
+    justify-content: center;
+    padding-block-start: ${pxToRem(80)};
+    padding-block-end: ${pxToRem(80)};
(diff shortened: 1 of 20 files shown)


packages/frontend/components/composed/issue/PrevNextIssue/PrevNextIssue.styles.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const FixedWidthButton = styled(Button)`
+  ${aButton("primary", "lg")}
+  width: ${pxToRem(180)};
+  justify-content: center;
+
+  & > svg {
+    flex-grow: 0;
+  }
+`;
```
1aurend, line 18: @jen-castiron I created this styled component to ensure that the Prev and Next buttons are the same size. Should we add this as a pattern to `Button` or just leave here for now?

jen-castiron: Let's leave this here for now - we might run into issues with mobile devices, and we can check when we work on the mobile styling

packages/frontend/components/composed/issue/PrevNextIssue/PrevNextIssue.styles.ts
```diff
@@ -0,0 +1,26 @@
+import styled from "styled-components";
+import { pxToRem } from "@wdp/lib/theme/functions";
+import { Button } from "components/atomic";
+import { aButton } from "theme/mixins/appearance";
+
+export const Inner = styled.div`
+  padding-block: var(--container-padding-lg);
```
jen-castiron, line 7: Safari doesn't recognize this as a property - for right now, `padding-block-start` and `padding-block-end` need to be defined. Same goes for any padding or margin inline. :(

jen-castiron: Oh - I think Dana came up with a mixin to solve for this in another project, maybe OFB?

PR #311, opened 2021-12-17
[F] Implement mobile navigation

1aurend:
Implements `MobileMenu` in `AppHeader` and `CommunityCondensedNav`. Needs accordion component (wdp-464) and updates to footer (wdp-462).


packages/frontend/components/global/AppHeader/AppHeader.styles.ts
```diff
@@ -1,25 +1,73 @@
...
 `;
 
 export const HeaderInner = styled.div`
   display: flex;
   align-items: center;
   justify-content: space-between;
+
+  ${globalNavRespond(
+    `
+    && {
```
jen-castiron, line 29: `&&` typo?

packages/frontend/components/layout/MobileMenuToggle/MobileMenuToggle.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import * as Styled from "./MobileMenuToggle.styles";
+import { IconFactory } from "components/factories";
+type IconProps = React.ComponentProps<typeof IconFactory>;
+
+export default function MobileMenuToggle({ onToggle, label, icon }: Props) {
+  return (
+    <Styled.Toggle onClick={onToggle}>
```
jen-castiron, line 8: Unless I missed it, it looks like we're missing the a11y attribute `aria-controls` for this toggle https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

packages/frontend/theme/mixins/base.ts
```diff
@@ -29,3 +29,9 @@ export function fluidScale(
   if (max === min) return max;
   return fluidScaleBase(max, min, defaultMaxVw, defaultMinVw);
 }
+
+export function globalNavRespond(content: CssContent, operator?: string) {
+  const breakpoint = breakpoints[70];
+
+  return respond(content, breakpoint, operator);
```
jen-castiron, line 36: nit? I think `breakpoint` can be defined as `70` here, since the respond mixin already looks for that breakpoint

jen-castiron (commented):
👏 👏 👏 

I found a few tiny style bugs, but otherwise it's looking really good.

There's a weird issue at exactly 768px width where the nav is missing, maybe because of the "min" operator:
Screen Shot 2021-12-17 at 12 09 44 PM

The nav buttons (picker and menu) are slightly off if the menu is opened or closed. (I might have introduced this to your code with the fluid scale for the containers left & right padding... the fluidscale should stay, but the nav might need some adjustment.) Here's a screenshot with the nav at 50% opacity so you can see what I mean:
Screen Shot 2021-12-17 at 12 12 23 PM

The navigation is a little bit smaller than in design (255px):
Screen Shot 2021-12-17 at 12 15 50 PM
Screen Shot 2021-12-17 at 12 16 32 PM

PR #319, opened 2021-12-20
[B] Split InstallationName and CommunityPicker and implement on mobile

1aurend:


packages/frontend/components/global/AppFooter/AppFooter.styles.ts
```diff
@@ -72,3 +81,26 @@ export const CopyrightText = styled.p`
...
 `;
+
+export const InstallationDesktop = styled.div`
+  align-items: center;
+
+  > p {
+    padding-block-end: ${pxToRem(3)};
+  }
+
+  ${respond(`display: none`, 70)}
```
jen-castiron, line 92: The missing `;` might be the reason why this component still shows on mobile:
Screen Shot 2021-12-20 at 12 08 53 PM

jen-castiron (commented):
The InstallationDesktop is still showing on mobile, but once that's fixed go ahead and merge this in 👍

PR #322, opened 2021-12-21
[B] Close dropdown when clicking current community

1aurend:


packages/frontend/components/atomic/BaseDropdown/BaseDropdown.tsx
```diff
@@ -37,7 +38,8 @@ const BaseDropdown = React.forwardRef<HTMLButtonElement, Props>(
           role={undefined}
           aria-label={label}
         >
-          {children}
+          {typeof children === "function" &&
+            children({ onClick: closePopover })}
```
1aurend, line 42: @jen-castiron Will you double check my types on this before I merge? Thanks!

packages/frontend/components/atomic/Dropdown/Dropdown.tsx
```diff
@@ -8,18 +8,24 @@ type BaseProps = React.ComponentProps<typeof BaseDropdown>;
...
-    return (
-      <Styled.Item key={i}>{item && React.cloneElement(item)}</Styled.Item>
-    );
-  });
+  const submenu = (onClick: React.MouseEventHandler | undefined) => {
+    return menuItems.map((item, i) => {
+      if (item === null) return null;
+      return (
+        <Styled.Item key={i}>
+          {item && React.cloneElement(item, { onClick })}
```
jen-castiron: There was a small problem here where a child's click event handler was getting overwritten. So if you had a button in the menu, any on click event would be changed to the hide function. I moved the onClick to the dropdown wrapper 👍

packages/frontend/components/atomic/BaseDropdown/BaseDropdown.tsx
jen-castiron: This check works, but if the children was not a function, it wasn't rendered. I added a small fix. 👍

jen-castiron:
I found a few small issues and added tiny fixes & comments. I'll merge this in once it finishes deploying 👍

jen-castiron:
Side note - this might be a good issue to bring up at a frontendo confab. It comes up a few times, but fix is always a little tricky, especially for a11y requirements. Would be good to get this infront on Dana just to make sure we've covered our bases.

PR #325, opened 2021-12-22
[F] Add cover placeholder generator

zdavis:


zdavis:
I think this is done, although I expect I'll do some work with Lael to refine the algorithm that generates the cover. 

@jen-castiron, before this is merged, there is one thing I'd like you to take a look at. The `CoverPlaceholder` component returns an SVG and currently it's necessary to pass a width to it (see line 31 of the `CoverImage` component). I expect there's a more sophisticated way to responsively size the SVG—my approach is pretty naive. If you can make this better, please do!

You can see an example of generated covers alongside image covers here: 

https://wdp-frontend-5vtbvevf6-cast-iron.vercel.app/communities/692oQObcg76IDn6T1bNZiX3G91oXEQ

Thanks!

jen-castiron:
This is pretty sweet! I've made a few minor adjustments: 
- Adjusted the placeholder styles to make it more responsive
- The gray cover's colors are already defined in `themes/variables`, so I updated the component to use the previously defined vars.
- Rebased 😄

PR #327, opened 2022-01-03
[F] Build mobile accordion component

1aurend:
@jen-castiron Here's a first pass on the Accordion component. Right now the styles are tightly coupled to the component. I think it would be easy to split this out into an unstyled `BaseAccordion` and `Accordion` like we have for Dropdown, but I thought it might be better to wait on that until we implement it in  more than one place and see what we need.

### DIFF
--- a/packages/frontend/components/atomic/Accordion/Accordion.stories.tsx
+++ b/packages/frontend/components/atomic/Accordion/Accordion.stories.tsx
@@ -1,0 +1,27 @@
+import { Story } from "@storybook/react";
+import Accordion from "./Accordion";
+
+type Props = React.ComponentProps<typeof Accordion>;
+
+export default {
+  title: "Components/Atomic/Accordion",
+  component: Accordion,
+};
+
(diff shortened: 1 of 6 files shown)


1aurend:
One other question: The `details` element maintains its open/close state if you close and then re-open the whole mobile menu. Is that expected, or should I look into closing the accordions when the menu closes?

jen-castiron:
I'm still not sure if the details/summary components should be used in a navigation menu... @dananjohnson , what do you think? It might be better (a11y wise) to make a custom disclosure, like this one: https://adrianroselli.com/2019/06/link-disclosure-widget-navigation.html or Reakit's disclosure: https://reakit.io/docs/disclosure/

jen-castiron:
To answer your question - yes, if `details` is open it will remain open.

dananjohnson:
Interesting. I can't think of a specific a11y issue with using the `details` element for a menu, but it does seem unconventional. FWIW this is how we handled mobile submenus on OFB, though I can't quite tell if that's how you're using it here, or for the open/close state of the whole menu.

PR #329, opened 2022-01-05
[F] Build browse schema route

jen-castiron:
This is turning out to be more of a bear than I thought...

- Adds browse route (may need to change name later...)
- Adds atomic Pagination component
- Adds atomic PageCount component

Example Issue order: https://wdp-frontend-git-jw-wdp-461-cast-iron.vercel.app/collections/gXy7BRxRH3MBniKd8LI7KmBfVr08EoMmG/browse/articles?page=1
Example Journal order: 
https://wdp-frontend-git-jw-wdp-461-cast-iron.vercel.app/collections/LvMzrv0hnkPTxevKFkekgHpXVR5y1bJ/browse/articles

TO DO:
~- [ ] Add browse page to items route~ I'm not convinced we need this yet
- [x] Add generic list item and styles to browse list
- [x] Consult folks on if the article and issue should follow style conventions in other lists
- [x] Remove current issue and featured issues from browse page
- [x] Update Issue ordering list to use same route / factory
- [x] Add pagination to Issue ordering list

To keep my sanity, I'm going to make these items a separate PR:
- Build and style grid
- Determine how the grid vs list view will be chosen. Ordering name?


zdavis:
Hey @jen-castiron in response to the last item in your list—"Determine how the grid vs list view will be chosen. We talked it over in a recent planning meetiong, but perhaps we should discuss this with Alexa. There's an issue—[WDP-459](https://linear.app/nglp/issue/WDP-459/declare-list-style-in-schemas)—that's related to this.

zdavis:
I do think that in the first iteration, at least, you could just pick a single list style (grid or list) and use that until the API gives you what you need to pick a more appropriate style.

PR #334, opened 2022-01-06
Ld/fe bugs

1aurend:
Resolves:

- wdp-473
- wdp-475
- wdp-479
- wdp-485
- wdp-487
- wdp-494
- wdp-496
- wdp-502


packages/frontend/components/composed/journal/RelatedJournals/RelatedJournals.tsx
```diff
@@ -115,18 +124,21 @@ const fragment = graphql`
               storage
               ...CoverImageFragment
             }
-          }
-          ... on Item {
-            __typename
-            id
-            title
-            slug
-            published {
```
jen-castiron, line 124: Side note - Since this is related journals, I think we can safely say the target is a Collection (rather than spreading the same props to on Collection and on Item)

1aurend: Right— this PR should include that change.

jen-castiron: Oh! Thanks - I miss read the change 😅

packages/frontend/components/composed/issue/IssueHero/IssueHero.tsx
```diff
@@ -69,10 +69,16 @@ export default function IssueHero({ data }: Props) {
...
+              {issue.issn && (
+                <Styled.MetaItem className="t-label-sm">
+                  ISSN: <Styled.Number>{issue.issn}</Styled.Number>
+                </Styled.MetaItem>
+              )}
               {issue.doi && (
-                <span className="t-label-sm">
-                  DOI: <Styled.Number>{issue.doi}</Styled.Number>
-                </span>
+                <Styled.MetaItem className="t-label-sm">
```
jen-castiron, line 78: nit: if this should be have as a block (see item styling), then I suggest replacing this Styled.MetaItem with a plain div.

packages/frontend/components/composed/journal/CurrentIssue/CurrentIssue.styles.ts
```diff
@@ -59,6 +63,8 @@ export const TitleBlock = styled.div`
   }
 `;
 
+export const Subtitle = styled.div``;
```
jen-castiron, line 66: nit: Same for Subtitle - if the tag doesn't have any styles, let's not define them here.

packages/frontend/components/composed/journal/CurrentIssue/CurrentIssue.tsx
```diff
@@ -54,6 +57,9 @@ export default function CurrentIssue({ data }: Props) {
                 <a>{issue.title}</a>
               </NamedLink>
             </h3>
+            <Styled.Subtitle className="t-copy-lighter t-copy-italic">
+              {issue.subtitle}
```
jen-castiron, line 61: There might not be a subtitle - check for a value before adding the div

packages/frontend/components/composed/journal/FeaturedJournals/FeaturedJournal.tsx
```diff
@@ -40,6 +40,9 @@ export default function FeaturedJournal({ data, index, coverHeight }: Props) {
       >
         <Styled.ItemText as="a">
           <Styled.ItemTitle>{journal.title}</Styled.ItemTitle>
+          <Styled.ItemSubtitle className="t-copy">
+            {journal.subtitle}
```
jen-castiron, line 44: Same here - subtitles aren't supposed to be required, so check before adding the div

PR #344, opened 2022-01-11
[F] Add HowToCite block

1aurend:
Can use [items/00GNkbL2fKp9uv8ErT8e1QTQzEwER47y](https://wdp-frontend-dr7i8gyc0-cast-iron.vercel.app/items/00GNkbL2fKp9uv8ErT8e1QTQzEwER47y) to view the block.


packages/frontend/components/composed/article/HowToCite/HowToCite.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const [plainText, setPlainText] = useState("");
+
+  useEffect(() => {
+    remark()
+      .use(strip)
+      .process(citation || "")
+      .then((value) => {
+        setPlainText(String(value));
+      });
+  }, [citation]);
```
1aurend, line 25: @jen-castiron Do you have a better way to strip the markdown for copying? I don't think this preserves bold/italics, so it's not ideal.

jen-castiron: How about using a ref on the ReactMarkdown element, and then when the button is clicked, copy the ref element's content? That should preserve styling and links.

packages/frontend/components/svgs/icons/Copy.tsx
```diff
@@ -0,0 +1,29 @@
...
+      xmlns="http://www.w3.org/2000/svg"
+      role="img"
+      {...props}
+    >
+      {title && <title>{title}</title>}
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M4.5 11.5H15V1H4.5V11.5ZM5.5 10.5V2H14V10.5H5.5Z"
+        fill="#232324"
```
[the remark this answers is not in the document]
1aurend: Thanks for the catch!

packages/frontend/components/composed/article/HowToCite/HowToCite.tsx
1aurend: I tried that first. `ReactMarkdown` doesn't accept a ref. 😞

jen-castiron: You can always wrap the `ReactMarkdown` component in a div and use that as a ref! I'm doing something similar in ArticleText, in order to get all of the headers after the text has rendered.

PR #354, opened 2022-01-12
[F] Build RelatedIssues block

1aurend:
View on `collections/84RDmoqs1BaJUZQdu2ypwI11Lbyy0Dp/browse/articles`


packages/frontend/components/composed/issue/RelatedIssues/RelatedIssueBlock.tsx
```diff
@@ -0,0 +1,86 @@
...
+            route={getRouteByEntityType(issue.__typename) || "home"}
+            routeParams={{ slug: issue.slug }}
+            passHref
+          >
+            <a>{issue.title}</a>
+          </NamedLink>
+        </h4>
+        <div className="t-copy-sm a-color-lighter">
+          {issue.volume && <p>{issue.volume.title}</p>}
+          {issue.journal && <p>{issue.journal.title}</p>}
```
1aurend, line 44: I threw this in in case there are linked issues not from the same journal, but can take out if all links will be from the same journal (or if just not helpful).

packages/frontend/components/composed/issue/RelatedIssues/RelatedIssueBlock.tsx
```diff
@@ -0,0 +1,86 @@
...
+import * as Styled from "./RelatedIssues.styles";
+import { NamedLink, CoverImage, PrecisionDate } from "components/atomic";
+import { RelatedIssueBlockFragment$key } from "@/relay/RelatedIssueBlockFragment.graphql";
+import { getRouteByEntityType } from "helpers";
+
+export default function RelatedIssueBlock({ data }: Props) {
+  const issue = useMaybeFragment(fragment, data);
+
+  return issue && issue.slug ? (
+    <li key={issue.slug}>
```
jen-castiron, line 13: nit: The `key` only needs to be defined here: https://github.com/NGLPteam/wdp-client/pull/354/files#diff-ec5b51868bbb5499cf5c1cfb4b9c6ec742cfcde2aa4d351700ec9fec461fedb4R23 , and can be removed from this component. 

Since you've been working on this project and OFB, which way do you prefer to structure lists and item components? 
```
# parent component
<ul>
   <li key={}><RelatedIssueBlock /></li>
</ul>
```
or 
```
<ul>
  <RelatedIssueItem key={} />
</ul>
```

I personally find it easier to keep the `li` tag with the `ul` tag, especially during the early stages of development. What do you think?

jen-castiron: @dananjohnson , I'm curious if you have a preference too ☝️

packages/frontend/components/composed/issue/RelatedIssues/RelatedIssueBlock.tsx
jen-castiron: That's a good question - I think issues would usually be from the same journal, so we wouldn't need the journal name. But my assumption might be wrong... @zdavis , what do you think?

jen-castiron:
Looking good! There's a little bit of metadata missing - Here's the metadata that needs to show for a list of issues (from WDP-518):
- issue title, subtitle, volume, published date, article count

packages/frontend/components/composed/issue/RelatedIssues/RelatedIssueBlock.tsx
zdavis: Pretty hard to say. Might have to see how this plays out in practice.

PR #363, opened 2022-01-14
[F] Add announcement pages

1aurend:
Connects announcements block to data and adds route for announcements: `/collections/[slug]/announcements/[announcement]`


jen-castiron (approved):
Love it, ship it 🚢

PR #366, opened 2022-01-14
[F] Add Research Units block

1aurend:


packages/frontend/components/composed/community/ResearchUnitsList/ResearchUnitsList.styles.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+`;
+
+export const UnitLink = styled.a`
+  display: flex;
+
+  &:hover,
+  &:focus {
+    text-decoration: underline;
+  }
```
1aurend, line 57: Is this right?

packages/frontend/components/composed/community/ResearchUnitsList/ResearchUnitsList.styles.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  ${respond(`min-height: ${pxToRem(200)};`, 70)}
+`;
+
+export const Image = styled(BaseImage)`
+  z-index: 0;
+  width: 100%;
+  height: auto;
+  border-radius: ${pxToRem(16)};
+`;
```
1aurend, line 94: I wasn't sure how to handle height here. This works okay with the placeholder image, but I wasn't sure if we wanted to allow for images of variable aspect ratios.

packages/frontend/components/composed/community/ResearchUnitsList/ResearchUnitsList.styles.tsx
jen-castiron: The link hover styles are a fancy box-shadow - if you use the `a-link` class, it will apply all the styles you need. I also suggest using the ArrowList component for the list

packages/frontend/components/composed/community/ResearchUnitsList/ResearchUnitsList.styles.tsx
jen-castiron: Use the height in the designs for now (which I think is what you're doing) - we can adjust when we have image data from the backend 👍

packages/frontend/components/composed/community/ResearchUnitsList/ResearchUnitsList.styles.tsx
```diff
@@ -0,0 +1,94 @@
...
+    `
+      grid-area: text;
+    `,
+    70
+  )}
+`;
+
+export const HeaderBlock = styled.div`
+  padding-block-end: ${pxToRem(12)};
+  border-bottom: 1px solid currentColor;
```
jen-castiron, line 36: There's a root variable `--border-color` that should match design and changes based on the background!

jen-castiron:
Sorry, last thing... The components in `composed` are organized by entity and schema. If these are meant for just units, it should be in a folder called `unit`

PR #379, opened 2022-01-19
[F] Add horizontal rule to form fields following groups

1aurend:


packages/admin/components/forms/FormGrid/FormGrid.styles.ts
```diff
@@ -21,4 +22,19 @@ export const Grid = styled.div`
...
+    margin-block-start: calc(${pxToRem(40)} - var(--form-grid-row-gap));
+
+    ${respond(
+      `
+        padding-block-start: ${pxToRem(48)};
+        margin-block-start: calc(${pxToRem(48)} - var(--form-grid-row-gap));
+
+      `,
+      50
+    )}
```
1aurend, line 38: @jen-castiron is there a better way to handle to specific spacing here? See: https://www.figma.com/file/EeaBT8NWvguKGhMQ7pgpry/NGLP-Admin-UI-Design?node-id=2007%3A11733

packages/admin/components/api/SchemaInstanceForm/Factories/GroupProperty/GroupProperty.tsx
```diff
@@ -18,13 +18,7 @@ export default function GroupProperty(props: Props) {
...
       </FormGrid>
     </FieldsetSection>
-  ) : (
-    <>
-      {group.properties.map((prop, index) => (
-        <SchemaProperty field={prop} key={index} />
-      ))}
-    </>
-  );
+  ) : null;
```
jen-castiron, line 21: Let's still show the group, maybe with a default legend "Fields". Otherwise we might not realize that's why the fields aren't showing in the admin app

PR #381, opened 2022-01-20
[F] Add app settings form, apply to frontend

jen-castiron:
To do:
- [ ] Determine if the "provider name" is the same as the institution name, or if we need a separate field
- [x] Fix flash of default theme
- [x] Check if user has permission to change global settings (globalAdmin)
- [x] nit: Consistent names (global settings vs app settings)
- [x] API: Add "gray" to list of possible colors


zdavis:
Added "gray" to API in https://github.com/NGLPteam/wdp-api/pull/35

PR #382, opened 2022-01-20
[C] Add test public pdf file, fix PDF issues

jen-castiron:


packages/frontend/components/composed/article/ArticleText/ArticleText.tsx
```diff
@@ -6,7 +6,14 @@ import * as Styled from "./ArticleText.styles";
...
 import { ArticleTextFragment$key } from "@/relay/ArticleTextFragment.graphql";
-import AssetInlinePDF from "components/composed/asset/AssetInlinePDF";
+// import AssetInlinePDF from "components/composed/asset/AssetInlinePDF";
+
+import dynamic from "next/dynamic";
+
+const AssetInlinePDF = dynamic(
+  () => import("components/composed/asset/AssetInlinePDF"),
+  { ssr: false }
+);
```
jen-castiron, line 16: @1aurend - After fussing with this and doing some digging, this code was the missing piece. I also added a isMounted check before rendering the react-pdf component. https://github.com/wojtekmaj/react-pdf/issues/136#issuecomment-716643894

The rest was just me playing around with prev/next buttons to make sure it worked 😅

jen-castiron:
@1aurend There's still a CORS issue that needs to get sorted out, so I tested this with a local file for now. (see public/pdf/Titanic.pdf)

packages/frontend/components/composed/article/ArticleText/ArticleText.tsx
1aurend: Thanks for figuring this out and sharing!

PR #387, opened 2022-01-24
[B] Set default to visible for new entities

1aurend:


jen-castiron:
@1aurend Let's bump the visibility field up on Collection/Item Edit forms as well, so it's in the same place for both new and edit forms

PR #388, opened 2022-01-24
[F] Save local user preference for opening new entities

1aurend:


packages/admin/components/composed/collection/CollectionAddForm/CollectionAddForm.tsx
```diff
@@ -22,7 +23,11 @@ export default function AddCollectionForm({
   data,
   parentId,
 }: Props) {
-  const [redirectOnSuccess, setRedirectOnSuccess] = useState(true);
+  const [prevRedirectState, setPrevRedirect] = useLocalStorage(
+    "nglp::open_collection",
```
jen-castiron, line 27: @1aurend I think you'll want this to be the same between collection and item forms, so any "Open ___ on save" checkbox is the same state as the last one. Could call it `nglp::open_entity_on_save` maybe?

PR #391, opened 2022-01-25
[F] Build initial parent selector modal

1aurend:
@jen-castiron I think this is as far as I can get on this until we have the api. Not sure whether we want to merge this in as is.


packages/admin/components/forms/ParentSelector/ParentSelectorModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+    viewer {
+      collections {
+        edges {
+          node {
+            id
+            title
+          }
+        }
+      }
+    }
```
1aurend, line 118: Right now, this doesn't return anything like on the top-level communities page. Same for items, so I didn't bother including both. Plus, we probably ultimately want a different list.

packages/admin/locales/en.json
```diff
@@ -354,6 +354,16 @@
         "change_label": "Change Schema",
         "change_message": "Are you sure you want to change this schema?"
       },
+      "parent": {
+        "label": "Parent",
+        "none": "No schema defined",
```
1aurend, line 359: Oops! Missed this one. 🙃

packages/admin/components/forms/ParentSelector/ParentSelectorModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+    communities {
+      edges {
+        node {
+          id
+          title
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
jen-castiron: For right now, use this communities list that you already have, and get the descendants on each node:
```
descendants {
        nodes {
          descendant {
            ... on Entity {
              title
            }
          }
        }
      }
```
There's a separate ticket for dealing with entity ambiguity. Bonus points for filtering by scope based on entity type, or schema based on parent schemas!
```
descendants(schema: ["nglp:journal_volume", "nglp:journal"], scope: COLLECTION)
```

packages/admin/components/forms/ParentSelector/ParentSelectorModal.tsx
jen-castiron: Since this returns nothing right now, let's get rid of it 👍

packages/admin/components/forms/ParentSelector/ParentSelector.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export default ParentSelector;
+
+type Props = {
+  data?: ParentSelectorFragment$key;
+};
+
+const fragment = graphql`
+  fragment ParentSelectorFragment on AnyEntity {
+    ... on Collection {
```
jen-castiron, line 48: If you want to try filtering by parent schemas, you might be able to get that data here and pass it to the ParentSelectorModal:
```
namedAncestors {
      ancestor {
        ... on Entity {
          title
          schemaVersion {
            slug
          }
        }
      }
    }
```

jen-castiron (commented):
Let's get something in the list for now, and not worry too much about entity ambiguity since it's in a separate ticket.

packages/admin/components/forms/ParentSelector/ParentSelector.tsx
1aurend: I decided this was probably too much, since this should ultimately come from the api. I did scope the query to right entity types.

packages/admin/components/forms/ParentSelector/ParentSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+import { ParentSelectorFragment$key } from "@/relay/ParentSelectorFragment.graphql";
+
+const ParentSelector = ({ data }: Props) => {
+  const entity = useMaybeFragment(fragment, data);
+  const parentFragment =
+    entity?.__typename === "Collection"
+      ? collectionParentFragment
+      : itemParentFragment;
+  const parent = useMaybeFragment(parentFragment, entity?.parent);
```
1aurend, line 19: @jen-castiron I couldn't figure out the type error here. I think I must be missing something simple. I tried splitting out the parent fragments to fix it, and that didn't work. Once this is resolved, this PR should be ready to go.

packages/admin/components/forms/ParentSelector/ParentSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { ButtonControl } from "components/atomic";
+
+import { ParentSelectorFragment$key } from "@/relay/ParentSelectorFragment.graphql";
+
+const ParentSelector = ({ data }: Props) => {
+  const entity = useMaybeFragment(fragment, data);
+  const parentFragment =
+    entity?.__typename === "Collection"
+      ? collectionParentFragment
+      : itemParentFragment;
```
jen-castiron, line 18: @1aurend Try fixing the type issue by checking for __typename === "Item", else return null. The error says that `parent` doesn't exist on __typename "%other", so we need to check that the typename is what we expect before getting the parent property.

PR #406, opened 2022-02-04
[Discussion][F] Filter schema select and add item/collection buttons based on schema ranks

jen-castiron:
*The problem:*
Users can add new items or collections with invalid schemas. This could lead to frustration on the user side, since the option for an invalid schema child was given. If we filter schemas, there's potential that the user could try to add an item and no schema type is available (Units, Series, and Campus, for example)

*The solution:*
Filter the schema select by the parent's schema ranks. This will give us a good idea of what kind of schemas are allowed on the parent. Check if the schema ranks include kind = ITEM or kind = COLLECTION before showing their relative add buttons.

*Example:* 
UC Merced is a Campus, and a Campus' schema ranks do not include any item schemas. The "Add Item" button is removed, because we need at least one valid schema type to be able to add an item:
Screen Shot 2022-02-04 at 1 57 27 PM

When adding a collection to this Campus, the schemas are filtered by the Campus' schema ranks:
Screen Shot 2022-02-04 at 1 58 39 PM

It was a pretty strait forward check, but any suggestions for API changes or other UX improvements are welcome.


jen-castiron:
There's one last problem to this: The "Items" tab navigation is still available, even though this particular collection (Campus) doesn't have any valid item schemas for children.

PR #421, opened 2022-02-16
[F] Add dashboard

jen-castiron:
Current list of collections is empty. Don't merge until API returns list of collections across communities.


scryptmouse:
By "collections across communities", do you mean all collections the user has access to? (aka `viewer.collections`). If so, that'll be out very shortly.

jen-castiron:
Yep! That's good to hear, thank you Alexa 😄

PR #461, opened 2022-04-19
[B] Handle html characters in contributor names

1aurend:


jen-castiron (approved):
Looks good, I didn't see any display issues with the new span tag 👍

PR #464, opened 2022-04-19
[B] Admin UI Fixes

1aurend:
Also resolves WDP-677

### DIFF
--- a/packages/admin/components/composed/item/ItemLayout/ItemLayout.tsx
+++ b/packages/admin/components/composed/item/ItemLayout/ItemLayout.tsx
@@ -47,7 +47,7 @@ export default function ItemLayout({
           memoizedItem.title || "glossary.item"
         );
         hideDialog();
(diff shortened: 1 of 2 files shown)


packages/admin/components/composed/item/ItemLayout/ItemLayout.tsx
```diff
@@ -47,7 +47,7 @@ export default function ItemLayout({
           memoizedItem.title || "glossary.item"
         );
         hideDialog();
-        router.replace(breadcrumbs[breadcrumbs.length - 1]?.href);
+        router.replace(breadcrumbs[breadcrumbs.length - 2]?.href);
```
jen-castiron, line 50: What was the bug with length-1?

1aurend: `breadcrumbs[length-1]` points to the item itself. We want the parent here because this is the redirect after the item is deleted. I think?

jen-castiron (approved):
🚢 🎊

PR #468, opened 2022-04-21
[F] Add entity scoping to search modal

jen-castiron:
- Resolves WDP-681


jen-castiron:
@1aurend I hear you, but I think we'll have to see if we get the same feedback from user testing at this point. The search is done inside of the entity you're looking at, so scoping up a level might be tricky. It's not the same as a filter, and might get a little confusing if you're under an item's header but you're searching the entire community. It's not impossible though. Would be good to bring up at our next NGLP moot.

PR #472, opened 2022-04-27
[F] Improve image loading UX

jen-castiron:
- Resolves WDP-668


jen-castiron:
There's currently an API issue with the new `hero` image size - merge this in when the API is fixtos

PR #476, opened 2022-05-03
[WIP][F] Fetch meta tag data server side

jen-castiron:
WIP.
Currently meta tag data is fetched client side. Some services do not wait for javascript to run before getting meta tag data, so the information could be lost. This is a test of getting the data server side using getStaticProps.

Meta tags generated on the homepage:
Screen Shot 2022-05-03 at 12 23 02 PM

### DIFF
--- a/packages/frontend/pages/index.tsx
+++ b/packages/frontend/pages/index.tsx
@@ -2,16 +2,38 @@ import React from "react";
 import { graphql } from "react-relay";
 import { QueryWrapper } from "@wdp/lib/api/components";
 import { useRouteSlug } from "@wdp/lib/routes";
+import { fetchQuery } from "relay-runtime";
+import { environment } from "@wdp/lib/app";
 import InstanceContentLayout from "components/composed/instance/InstanceContentLayout";
 import { pagesQuery as Query } from "@/relay/pagesQuery.graphql";
-
+import {
+  pagesIndexQuery,
+  pagesIndexQueryResponse,
+} from "@/relay/pagesIndexQuery.graphql";
 import AppLayout from "components/global/AppLayout";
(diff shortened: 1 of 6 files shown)


packages/frontend/pages/index.tsx
```diff
@@ -2,16 +2,38 @@ import React from "react";
...
-
+import {
+  pagesIndexQuery,
+  pagesIndexQueryResponse,
+} from "@/relay/pagesIndexQuery.graphql";
 import AppLayout from "components/global/AppLayout";
+import { AppHtmlHead } from "components/global";
+
+export async function getStaticProps() {
+  const env = environment();
```
jen-castiron, line 17: It seems strange to me that I have to make a new environment here. Since this data is on the frontend and would not require a sign in, I'm skipping the keycloak token.

packages/frontend/pages/index.tsx
```diff
@@ -2,16 +2,38 @@ import React from "react";
...
+  };
+}
 
-export default function HomePage() {
+export default function HomePage(props: pagesIndexQueryResponse) {
   const slug = useRouteSlug();
 
   return (
     <>
+      <AppHtmlHead data={props} />
```
jen-castiron, line 36: This will need to be included on every page, but for now it's here as a test.

packages/lib/app/environment.ts
```diff
@@ -13,7 +13,7 @@ interface KeycloakRef {
 }
 
 export default function buildEnvironment(
-  keycloakRef: KeycloakRef,
+  keycloakRef?: KeycloakRef,
```
jen-castiron, line 16: Making keycloak ref optional for environments in getStaticProps. Open to suggestions! I could be doing this incorrectly.

PR #479, opened 2022-05-24
[F] Build entity selector

1aurend:
@jen-castiron Here's an initial version of the selector. I've implemented it first in `ParentSelectorModal` like you suggested.

I'd love any feedback you have, but in particular, I think there's work to do on the UX of selecting an entity versus navigating into it. For instance, the link with the name of the entity— should that open the children or select the entity? 

In terms of function, I've added a couple of props, `startEntity` and `scopeToCommunity`. The former opens the selector at a particular entity in the hierarchy; the latter makes it so a user can only select entities within a given community. I don't know if we need this, but it seemed to make sense for reparenting, at least in some cases. I don't think it would be hard to expand the scope to functionality to collections also, if that is desirable.

I also ran into some issues integrating the selector with the form. I had to add `e.preventDefault` to all of my click handlers in order to avoid triggering form submit when navigating through the selector. This may be fine, but maybe you see a better solution.


1aurend:
Just a note for myself to add an option `omitSelf` from this list of entities.

1aurend:
@jen-castiron This is ready for review! I love the hidden radio look. ✨ See my comments on the props of `Controller` for config options.

There's one remaining issue, which isn't about the entity selector per se, but form validation doesn't work for the parent selector modal (and wasn't working previously). The error names returned from the server don't match the form field name `parentId` so the error component we use from react-hook-form doesn't render anything. I have an idea for how to fix this, since we don't want to show the server error text regardless, but I'll make a separate ticket for it, so I can run it by you.

jen-castiron:
@1aurend A couple small things:
* The text should line up with the parent text
* The save button is cut off in small browsers / devices, and the buttons could be cut off and unreachable
Screen Shot 2022-06-07 at 2 25 42 PM

Screen Shot 2022-06-07 at 2 26 37 PM

1aurend:
Thanks for the catch on the padding @jen-castiron. I must have messed it up when I was adjusting the clickable areas. Re: small screens— I adjusted the height of the selector to use a `vh` unit rather than `px`. Can you think of a better way to set the height?

PR #481, opened 2022-05-24
[B] Fix item and collection layout loading issue

jen-castiron:
WIP

Current behavior:
Clicking tabs on item pages bumps user back to the top of the page. 

Expected behavior: 
Clicking tabs should only load the new page, and hero elements should not reload.

Technical Issues:
Each page runs a query that gets the layout data, including any header elements that appear on each child page. The item layout fragment isn't caching, or isn't caching in a way we expect. We can either:
1. Separate queries for the app/item layout and any child routes (pages, browse, etc)
2. Fix the caching issue 

It's preferable if we fix the caching issue and make any long term improvements to the fragments rather than split queries. The backend currently gets around this by using `getLayout`. This fix attempts to use `getLayout` effectively in the frontend.


jen-castiron:
@1aurend This is a huuuuuge update, so the review here is for testing the frontend preview for any regressions. 

Some of the major changes:
- Wrapped item and collection routes in the layouts `CollectionLayoutQuery` and `ItemLayoutQuery`
- Moved EntityLayoutFactory from other layout factories into `CollectionLayoutQuery` and `ItemLayoutQuery`. This is so heros and bar navigation doesn't reload when the sub route changes
- Changed the page loading to only show a spinner if the LayoutQuery is loading. Child components take care of their own loading states

PR #482, opened 2022-05-24
[B] Fix scroll issue in items and collections nav

jen-castiron:
- Resolves WDP-734


jen-castiron:
This change was no bueno

PR #485, opened 2022-05-25
[B] Add ORCID + update affiliation for article contributors

1aurend:


1aurend:
> @1aurend While reviewing this I found an ORCIDLink component I made a while back - can we use that one, or is there different styling for this link?

I also noticed it, but after I was almost done putting this together. I think the styling is slightly different because I thought the icon looked better darker than the text here. But happy to test the other component out if you like!

jen-castiron:
@1aurend Let's use the existing ORCID link component. I'm OK with changing the color of the icon to something darker - just check that it still looks good in the contributor detail page. 👍

1aurend:
@jen-castiron That was a ridiculously simple swap. I feel silly 🙃 .

packages/frontend/components/composed/contribution/ContributionsBlock/ContributionBlockItem/ContributionBlockItem.styles.ts
```diff
@@ -9,18 +10,32 @@ export const ListItem = styled.li`
...
+export const ItemAvatar = styled.a`
   flex: none;
 `;
 
 export const ItemMetadata = styled.div`
   padding-block-start: var(--padding-xs);
 `;
+
+export const ORCIDLink = styled(ORCIDLinkBase)`
+  ${aLink}
```
jen-castiron, line 30: Since the base includes link styling, this can be removed. Even with that change, the hover is a little odd:
Screen Shot 2022-05-26 at 11 26 49 AM
Because the fancy hover effect is a background, not underline text, the line appears at the bottom of the span block. It's tricky to get right!

Here's the styling with everything but the `svg >` block removed. It looks pretty good, and I'd be OK with it like this, but if we want the flex layout we need to do something like the below option.
Screen Shot 2022-05-26 at 11 32 25 AM

For the flex layout, try updating the atomic component to this, and then style the `a` tag with the flex layout. If we make the change here, we can keep the link display consistent between both pages.
```
/* Simple download text and icon,
 * style can be changed using the className property */
export default function ORCIDLink({ children, ...props }: BaseProps) {
  return children ? (
    // Change `a` to a styled link with flex
    <a target="_blank" {...props}>
      <IconFactory icon="orcid" />
      // The utility class a-link has link styling
      <span className="a-link">{children}</span>
    </a>
  ) : null;
}
```

zdavis:
I need to accept a PR to test Heroku github triggered deploys, and this looks to me to be done at this point.

PR #492, opened 2022-05-31
[B] A11y fixes for article PDF page

jen-castiron:


jen-castiron:
Hey @dananjohnson , can I get your eyes on this? 👁️ The article page "failed" on screen readers, according to this ticket: https://linear.app/nglp/issue/WDP-733/check-article-page-in-screen-reader.

> Lisa used https://assistivlabs.com/ to review item page and reports that it failed to detect inline PDF or any of the information in the tabs (e.g. metadata, contributors)

I added some text to the loading indicators, which might help. I tried adding an aria-live region around the PDF loader, but it became very noisy, saying "Loading page 1, loading page 5, loading page 10..." I'm not sure what to do here to make it more a11y friendly... what are your suggestions?

A good page for testing is https://wdp-frontend-git-jw-wdp-733-cast-iron.vercel.app/items/Yv3YGPBhqvnmI27Ybum7Nzf6xEgmnOKE

dananjohnson:
@jen-castiron you might try `aria-busy`, though browser/SR support seems limited. See https://adrianroselli.com/2020/11/more-accessible-skeletons.html. I'm also curious what they mean when they say that this tool "failed to detect" the content. What SR? What browsing mode? I don't have access to the Linear issue, so feel free to add me if there's more information there.

jen-castiron:
@dananjohnson I asked, but I never got a response. Closing this PR and adding some of the changes to the other PDF PR

PR #493, opened 2022-05-31
[B] Fix font loading issue

jen-castiron:
@dananjohnson I could use your help with this one too. The font "jiggles" when you change pages, and I'm not quite sure why. I've moved the global stylesheet out of the theme provider, which helped a little, but there's still a noticeable layout shift. It still happens if you remove the font setting on ThemeProvider (line 37 of AppContextProvider). Any help would be appreciated!


dananjohnson:
@jen-castiron it looks like the font files are being re-downloaded on every page change. This is probably why: `cache-control: public, max-age=0, must-revalidate`. Probably just need to tweak cache settings in Vercel. There's also some stuff you can do to style your fallback font while the primary one loads which can help with layout shift. But start with the cache policy change and see if that's good enough.

PR #494, opened 2022-05-31
[F] Update admin global nav design

1aurend:
@jen-castiron Could you take a look at the hover style for the installation name? Should it match the thick underline of the dropdowns or is default link styling okay here? Thanks!


jen-castiron:
@1aurend The hover looks OK to me, since it's the same hover used on that style of text. Let's leave it for now. I did notice that the padding on the left needs to be removed, and we should probably prevent flex wrapping on most mobile devices:
Screen Shot 2022-05-31 at 3 55 03 PM

PR #497, opened 2022-06-01
[F] Improve PDF UX, lazy load pages

jen-castiron:
What changed:
* PDF navigation lazy loads each page
* PDF pages load on scroll using an intersection observer. Pages within a certain threshold (200px) around the scrolled page should also load.

Known issues:
* PDF load is improved, but large PDFs still have some lag
* The first nav thumbnail is offset (easy fix, just need to clean up code)

### DIFF
--- a/packages/frontend/components/atomic/loading/LoadingBlock/LoadingBlock.tsx
+++ b/packages/frontend/components/atomic/loading/LoadingBlock/LoadingBlock.tsx
@@ -1,10 +1,22 @@
 import React from "react";
+import { useTranslation } from "react-i18next";
 import LoadingSpinner from "../LoadingSpinner";
 import * as Styled from "./LoadingBlock.styles";
 
 export default function LoadingBlock({ className, label, style }: Props) {
+  const { t } = useTranslation();
+
   return (
-    <Styled.Wrapper className={className} style={style}>
+    <Styled.Wrapper
+      className={className}
+      style={style}
+      aria-live="polite"
+      aria-busy="true"
+      aria-valuemin={0}
+      aria-valuemax={100}
+      role="progressbar"
+      aria-valuetext={label || t("common.loading")}
+    >
       <LoadingSpinner label={label} />
     </Styled.Wrapper>
   );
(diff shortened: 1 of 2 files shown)


jen-castiron:
@dananjohnson @1aurend This could use some testing, and some thoughts on improvements & a11y concerns. 

Simple example: items/Yv3YGPBhqvnmI27Ybum7Nzf6xEgmnOKE
Some PDFs load better than they did before, but they're still > 100 pages and a heavy load: items/rdMgPRnt0gXgIrL3coyDkCGyn9zg6vP

jen-castiron:
@zdavis It was suggested that we only show the first 25 pages of a PDF, then prompt the user to download the rest. What do you think?

zdavis:
seems reasonable to me.

dananjohnson:
@jen-castiron I like what you did with the 25-page limit. Seems reasonable not to load any more than that.

I took the page on a test drive with VoiceOver and my keyboard and felt like the structure could be made a bit more clear for ATs. Here are some suggestions:

* **Priority**
  * Heading structure should have an h1 and then descend in order. Currently, the page title is an h2 and there are two other headings on the page: an h4 and an h5.
  * Similarly, the tabbed sections (full text, metadata, etc) should have a hidden h2 element to make the page structure more clear and to allow easier navigation.
* **Nice-to-have**
  * The PDF thumbnail sidebar should have an `<ol>` list structure.
  * Adding an `aria-label` to the tab-looking `<nav>` element would make it more clear what it is.
  * I wonder if the Full Text section should have a skip-to link to the full PDF text? That would let people avoid tabbing through potentially 25 thumbnail links.

jen-castiron:
@dananjohnson Thank you so much! I implemented the changes you suggested, but I wasn't sure what to label the navigation area with the `h2` tag. Are we labelling that there's "Article Content" below, or that it's an "Article Navigation"?

dananjohnson:
Oh, I meant that each of the tabbed sections (full text, metadata) should have its own `h2`, maybe just with the same text as the tab title, and that the `nav` would have its own `aria-label`, like `Article subpages` or something.

```html
<nav aria-label="Article subpages">...</nav>
<section>
  <h2 class="a-hidden">Full text</h2>
  ...
</section>
```

Does that make sense?

jen-castiron:
@dananjohnson That makes total sense. I'll add it in. 👍

PR #502, opened 2022-06-07
[B] Update number formatting in PageCount

1aurend:
@jen-castiron is it okay that I updated i18next here?


jen-castiron:
@1aurend Unfortunately updating i18n caused major issues with storybook and admin last time I tried. The local admin no longer loaded, and storybook was having trouble building.

1aurend:
> 

Interesting. I didn't try loading storybook, but I was running admin at the same time I was working on this with no problem. I'll roll it back and see if we need the number formatting polyfill.

PR #507, opened 2022-06-08
[B] Shrink LoadingCircle in small containers

1aurend:


jen-castiron (changes requested):
This change had a funny side effect - The spinner for the admin pages is so tiny! 😆 
Screen Shot 2022-06-09 at 9 17 59 AM

PR #513, opened 2022-06-29
[F] Implement prefix search in typeaheads

1aurend:


packages/admin/components/forms/CollectionTypeahead/CollectionTypeahead.tsx
```diff
@@ -1,77 +1,105 @@
...
-          slug
+const query = graphql`
+  query CollectionTypeaheadQuery($query: String!) {
+    search(visibility: ALL) {
+      results(
+        prefix: $query
+        page: 1
+        perPage: 50
+        order: TITLE_ASCENDING
+        schema: [
```
jen-castiron, line 85: Do the schemas need to be hardcoded here?

packages/admin/components/forms/ItemTypeahead/ItemTypeahead.tsx
```diff
@@ -1,77 +1,103 @@
...
-      name={name}
-      control={control}
-      rules={{
-        validate: (value) => {
-          return !!value || (t("forms.validation.item") as string);
-        },
+  const debouncedOnChange = debounce((value) => setQ(value), 500);
+
+  return (
+    <QueryWrapper<Query> query={query} initialVariables={{ query: q }}>
```
jen-castiron, line 41: These changes work well, but I am concerned that the entire Controller and BaseTypeahead might be getting re-rendered with every query when just the list of options is changing. I would see if these components might benefit from using the `useQuery` hook with a memoized options list rather than the `QueryWrapper` component, if there's time.

packages/admin/components/forms/CollectionTypeahead/CollectionTypeahead.tsx
1aurend: Right — yes, that was one of my questions too. We can certainly pass them in as an argument, but how do we determine, for a given instance, what the full set of collection/item/etc. schema's is? That would be a prior query, right?

jen-castiron: I think I see... Is the issue how to search for just collections? I had to look at the Collections page to remember! You'll need to add the scope filter, `scope: COLLECTION` in this case. If we do need to filter down further by schemas, then we would need to make a prior query on the parent.

packages/admin/components/forms/ItemTypeahead/ItemTypeahead.tsx
jen-castiron: Hmm, I think you're right. My assumption was that moving the value to a hook and constant, rather than a wrapper component, would prevent Controller from re-rendering. I'm OK with leaving it as is and keeping an eye on performance issues.

packages/admin/components/forms/CollectionTypeahead/CollectionTypeahead.tsx
1aurend: I'm not sure that works because there isn't necessarily a parent, e.g. [here](https://wdp-admin-git-ld-wdp-750-cast-iron.vercel.app/users/GD7rv85CxX97UBONt8BoeSNbEdL2Xk6/collections?order=RECENT&drawer=addUserCollectionAccess&drawerSlug=GD7rv85CxX97UBONt8BoeSNbEdL2Xk6). And we don't want to limit it to direct descendants like `scope` does.

1aurend: Where are you looking at the Collections page? I think I'm missing something.

jen-castiron: I looked at `CollectionListSearchFragment` - `scope` isn't limited to direct descendants, unless you add `maxDepth: 1` to the search filters!

jen-castiron: (See `packages/admin/pages/collections/index.tsx`)

1aurend: Oh! The debug site confused me. It says: "This will include only directly descending Collections, no links."

1aurend: I will make this change then.

PR #514, opened 2022-06-30
[B] Fix retry condition in QueryWrapper

1aurend:
The answer to why opening drawers in admin reloaded the underlying content turned out to be a bug in the refetch logic in `QueryWrapper`:
```
useEffect(() => {
    if (!intersection(refetchTags, triggeredRefetchTags)) return;
    retry();
  }, [triggeredRefetchTags, refetchTags, retry]);
```
`!intersection(refetchTags, triggeredRefetchTags)` returns an empty array if there are no matches, so this condition was always true.

Also resolves:
- WDP-800


jen-castiron (approved):
Excellent find!

PR #518, opened 2022-07-01
[B] Fix body scroll lock for drawers

1aurend:
@jen-castiron I think this works, but it could definitely use another set of 👀 before merging. As I mentioned in Linear, even if this works, I think we could further improve performance of lists with lots of dialog components. It seems like body styles get set too many times by the various dialogs.

### DIFF
--- a/packages/admin/components/layout/Drawer/Drawer.tsx
+++ b/packages/admin/components/layout/Drawer/Drawer.tsx
@@ -7,6 +7,7 @@ import * as Styled from "./Drawer.styles";
 import { ButtonControl } from "components/atomic/buttons/";
 import { useQueryStateContext } from "hooks";
 import LoadingCircle from "components/atomic/loading/LoadingCircle";
+import usePreventBodyScroll from "components/layout/Modal/hooks/usePreventBodyScroll";
 
 /**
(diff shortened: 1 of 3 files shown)


packages/admin/components/layout/Drawer/Drawer.tsx
```diff
@@ -31,13 +32,16 @@ const Drawer = ({
...
+  usePreventBodyScroll(dialog.visible);
+
   return (
     <Styled.DialogBackdrop {...dialog}>
       <Styled.Dialog
         aria-labelledby={uidLabel}
         aria-describedby={uidDesc}
         hideOnClickOutside={hideOnClickOutside}
         {...dialog}
+        preventBodyScroll={false}
```
jen-castiron, line 44: @1aurend tiny nit - spread operators should be at the end of the list

jen-castiron (approved):
👍 Looks good to me. Can you make an issue with your thoughts on dialog improvements? That way we don't forget. 😄

packages/admin/components/layout/Drawer/Drawer.tsx
1aurend: Oh, yep. Np!

PR #520, opened 2022-07-05
[WIP][F] Add search to contributors list

1aurend:
Adds `prefix` search to contributors list, but probably requires an api update for a full search. Search should probably also be added to the users list.


zdavis:
@jen-castiron @1aurend this one has been hanging out for a while. Can it be merged?

1aurend:
I left it work in progress because prefix searching only allow for searching by first name. If you try to search by last name, you get no results. We don't have a general contributor or user search api. If it's okay to have first name search only, then this works and can be merged.

jen-castiron:
First name search is better than no search. @1aurend , I'll get this merged into the next branch so it can go in with the other updates. @zdavis , I'll make an issue for Alexa - this looks like the API needs to be updated to search for last names, so it shouldn't require more FE changes.

PR #522, opened 2022-07-06
[POC][F] Build EntitySelectorDisclosure

1aurend:
@jen-castiron I think the combo typehead + entity selector works! Let me know if you see performance or UX issues.

Can preview [here](https://wdp-admin-lrzkzz54f-cast-iron.vercel.app/contributors/J6GNobp5TD7yBUwZdt0DJ7C5zP7Q2azL/items?order=RECENT&drawer=addContributorContribution&drawerSlug=J6GNobp5TD7yBUwZdt0DJ7C5zP7Q2azL&drawerContributionType=item).

### DIFF
--- a/packages/admin/__generated__/EntityTypeaheadQuery.graphql.ts
+++ b/packages/admin/__generated__/EntityTypeaheadQuery.graphql.ts
@@ -1,0 +1,312 @@
+/* tslint:disable */
+/* eslint-disable */
+// @ts-nocheck
+
+import { ConcreteRequest } from "relay-runtime";
+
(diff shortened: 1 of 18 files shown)


jen-castiron:
@1aurend I'm having trouble using the typeahead - if I type too fast, it doesn't register all the key strokes. I'll see if I can find the root of the problem, because I really want to get this in!

1aurend:
> @1aurend I'm having trouble using the typeahead - if I type too fast, it doesn't register all the key strokes. I'll see if I can find the root of the problem, because I really want to get this in!

We did something on Scarecrow that improved the way state was handled, and I remember thinking it should get ported back over here. But, of course, I don't remember what now...

PR #527, opened 2022-07-26
[B] Redirect users if they land on a disabled ordering page

jen-castiron:
NGLP Team Publications has two orderings:
- Series 
- Units (disabled)

If the user tries to land on the disabled unit ordering, it should redirect to the collection home page.
https://wdp-frontend-8lzxgucru-cast-iron.vercel.app/collections/J6QX5VK5TOVzceMPc0VJMTan7nEZVOq/browse/units

None of the collection or item navigations should show any disabled orderings.

@zdavis , if you have a few minutes try going to the disabled unit above. I'm going to double check that this won't result in an infinite loop.

Edit: Updated with a check to make sure the initial ordering isn't disabled


zdavis:
I didn't get an infinite loop, FWIW.

PR #529, opened 2022-07-27
[F] Add missing refetch tags

jen-castiron:


jen-castiron:
Hey @zdavis ! I added refresh tags to CollectionLayoutQuery, and added refresh tags for announcements and orderings. Announcements works great, but orderings goes into an infinite loop. I haven't figured out why yet, and I thought you might want to take a look too.

zdavis:
Thanks! I’ll take a look tomorrow.

PR #534, opened 2022-07-27
[B] Filter ordering options

jen-castiron:


jen-castiron:
@zdavis The list of ordering options (`orderingPaths`) is long and difficult to parse. This PR filters this list into a select number of options.
```
entity.created_at
entity.published
entity.title
entity.updated_at
```
I've set this up so more options can be added per instance using an env var:
```
entity.depth
props.accessioned
props.available
props.meta.collected
props.id
props.volume.sortable_number
props.number
schema.kind
schema.name
```
Before (Journal):
Screen Shot 2022-07-27 at 3 03 34 PM

After:
Screen Shot 2022-07-27 at 3 03 58 PM


What do you think? Does it make sense to use an env var for different instances, or should it be a fixed constant? Is there anything you would add or remove from this list.

zdavis:
I like the idea of using an env var for now. I think the default should include, at the very least, the following:

entity.title
entity.published
entity.created_at
entity.updated_at
props.volume.sortable_number
props.sortable_number
props.id

These are the properties that show up in the schema-defined orderings, and I think we'd need to see them in the options as well, right?

jen-castiron:
That's true - I'll pop those in as defaults

zdavis:
Generally speaking, I think I'd be inclined to have the minimum number of options available here and then add them as it becomes clear people need them, rather than overwhelm them with a bunch of stuff.

PR #538, opened 2022-09-19
[F] Analytics

1aurend:


jen-castiron:
@1aurend Adjusted so the fixed size is the same as the original desktop size

jen-castiron (approved):
👍 Go ahead and merge when you're ready

PR #544, opened 2022-09-28
[F] Adjust map styling for mobile devices

jen-castiron:


jen-castiron:
@1aurend Adjusted so the fixed size is the same as the original desktop size

1aurend:
I like the map with the adjustments. If we go with the side scrolling should we also implement it for the chart, so it's consistent?

jen-castiron:
@1aurend Yeah, we probably should - I'll see if I can't wrap it up with the chart today.

PR #553, opened 2022-12-05
[F] Generate site maps

jen-castiron:


jen-castiron:
@zdavis This should satisfy the sitemap ask for google scholar. The site is broken down into several smaller sitemap indexes, and each of those either include a further breakdown of collection/item children, or lead to a sitemap with page urls. I can give you a small demo if you'd like

PR #577, opened 2024-07-30
[F] Add refetch tags to communities, collections, items, contributors list

jen-castiron:
- Resolves MERU-855

@1aurend is there a refetch mechanism that should be working instead of refetch tags? Or is it OK to add these refetch tags back in?


1aurend:
> * Resolves MERU-855
> 
> 
> @1aurend is there a refetch mechanism that should be working instead of refetch tags? Or is it OK to add these refetch tags back in?

Hmm... I'm not 100% sure, but I thought the refetch tags no longer worked with the refactored queries. There is a `subscribeIds` prop on the query wrappers:
```
return (
    <QueryTransitionWrapper<Query>
      query={query}
      variables={contributorSearchVars}
      subscribeIds={["Contributor"]}
      loadingFallback={<LoadingPage />}
    >
      {({ queryRef }) =>
        queryRef ? <ListQuery queryRef={queryRef} /> : <ContributorList />
      }
    </QueryTransitionWrapper>
  );
}
```
Might be worth a look why the subscribeId isn't working?

PR #579, opened 2024-08-07
[B] Update Meru frontend to next-auth

jen-castiron:


jen-castiron:
@1aurend I saw the package/auth folder had everything we need to use next-auth, but it wasn't working correctly when I tried importing it from that package. I copied over what was in package/auth into package/frontend/lib/auth, just to make sure it worked. There's one other issue - the Environment type isn't quite right in [RelayClientEnvProvider](https://github.com/NGLPteam/wdp-client/blob/af412a1ad09efaf23929243982e38e58273f9f21/packages/frontend/lib/relay/RelayClientEnvProvider.tsx#L29-L30). See the [Vercel build](https://vercel.com/cast-iron/wdp-frontend/8wZwYvnMwQV64a5ahWdahxEqoY5w?filter=errors) for the error

jen-castiron:
@scryptmouse Let me know when the env vars NEXTAUTH_URL (FE domain name) and
NEXTAUTH_SECRET secret key (generated string) are set on prod, and I'll merge this in! If there's a way I can add these just let me know.

PR #586, opened 2024-09-05
[C] Fix styled component DOM warnings, errors

jen-castiron:
After styled components was upgraded, styled-components started adding passed attributes to the DOM. The attributes need to be renamed with a leading `$` symbol. This change primarily helps developers - if the styled-component errors are removed, the more important errors are easier to parse.

### DIFF
--- a/packages/admin/components/atomic/navs/MobileSubNav/MobileSubNav.tsx
+++ b/packages/admin/components/atomic/navs/MobileSubNav/MobileSubNav.tsx
@@ -38,7 +38,9 @@ const MobileSubNav = ({ tabRoutes, sidebarLinks }: Props) => {
         ref={mobileNavRef}
       >
         <MobileMenuList>
-          {tabRoutes?.map(({ label = "", ...namedLinkProps }, i) => (
+          {/* Do not pass actions down to the DOM */}
+          {/* eslint-disable-next-line @typescript-eslint/no-unused-vars */}
+          {tabRoutes?.map(({ label = "", actions, ...namedLinkProps }, i) => (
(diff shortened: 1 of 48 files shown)


packages/admin/components/atomic/navs/MobileSubNav/MobileSubNav.tsx
```diff
@@ -38,7 +38,9 @@ const MobileSubNav = ({ tabRoutes, sidebarLinks }: Props) => {
         ref={mobileNavRef}
       >
         <MobileMenuList>
-          {tabRoutes?.map(({ label = "", ...namedLinkProps }, i) => (
+          {/* Do not pass actions down to the DOM */}
+          {/* eslint-disable-next-line @typescript-eslint/no-unused-vars */}
+          {tabRoutes?.map(({ label = "", actions, ...namedLinkProps }, i) => (
```
1aurend, line 43: What do you think about using an ignore pattern rather than disable comments? Something like...
```
// in rules in eslint.config.js

"@typescript-eslint/no-unused-vars": [
      "error",
      {
        varsIgnorePattern: "^_",
        argsIgnorePattern: "^_",
        destructuredArrayIgnorePattern: "^_",
        ignoreRestSiblings: true,
        caughtErrors: "none",
      },
    ],
```

1aurend (approved):
@jen-castiron code looks good to me! It's kind of fun looking back at this and seeing how much we've learned since. I realized I don't have a keycloak login for the new staging host, so I can't click around tonight, but I can do that tomorrow if you'd like.

PR #591, opened 2024-09-09
[B] Fix broken search on schema in admin

jen-castiron:
- Resolves MERU-901


jen-castiron:
Blocked - appears to also be a BE issue. See MERU-901

PR #596, opened 2024-09-12
[B] Add metadata to frontend pages

jen-castiron:
- Resolves MERU-913


1aurend (approved):
@jen-castiron looks good to me and doesn't duplicate anything I've done. Want to merge before I put in my updates from last week, so you don't have to rebase?

PR #601, opened 2024-12-05
[B] Entity page body fixes

1aurend:
Resolves MERU-926

### DIFF
--- a/packages/admin/components/api/MutationForm/MutationForm.tsx
+++ b/packages/admin/components/api/MutationForm/MutationForm.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo } from "react";
+import React, { useCallback, useMemo, useEffect } from "react";
 import { useMutation, graphql, GraphQLTaggedNode } from "react-relay";
 
 import { useForm, FormProvider } from "react-hook-form";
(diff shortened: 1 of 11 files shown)


packages/admin/components/api/MutationForm/MutationForm.tsx
```diff
@@ -251,6 +254,10 @@ export default function MutationForm<
     return loading || isSubmitting || isValidating;
   }, [form, loading]);
 
+  useEffect(() => {
+    form.reset(defaultValues);
+  }, [defaultValues]);
+
```
1aurend, line 260: @jen-castiron Do you have time to give this PR a quick review? I had to adjust both mutation form here and the fetch policy in lazy load query in order to not see stale data in the drawer. I'd appreciate a second set of eyes to make sure these changes don't break anything else or that there isn't a simpler solution I'm missing. Here's the original basecamp thread with my response to Sarah: https://3.basecamp.com/4006493/buckets/36220575/messages/7852803463

jen-castiron: I'll take a look!

jen-castiron: @1aurend There is a much simpler solution - add `body` to the fragment returned on update. That will trigger a relay cache update, and you should see the new value when you re-open the drawer. 
```
const mutation = graphql`
  mutation EntityPageUpdateFormMutation($input: UpdatePageInput!) {
    updatePage(input: $input) {
      page {
        id
        title
        slug
        position
        body
        ...
```
It's easy to forget to update the mutation - in some cases I'll make a fragment for default values, just so I only have to update it in one place

jen-castiron: Hokay - so. I pulled this branch down and ran `yarn relay` to update the fragment. Then I removed the useEffect from MutationForm. The body field is updating as expected, so, unless there's another reason to keep it, I think you can remove the code below after running `yarn relay`.
```
useEffect(() => {
    form.reset(defaultValues);
  }, [defaultValues]);
```

